PR #16, opened 2018-05-12
added feature where PR is not be mergeable when there is no description

### LINKED ISSUE #14: Prevent merge when there is no description
The PR should not be mergeable when there is no description on the PR. Suggested description:

"Description on the pull request should not be empty."

shine2lay:
This PR implements the features described in #14.

closes #14


lib/body.js
```diff
@@ -0,0 +1,21 @@
...
+ * @return
+ *  JSON object with the properties mergeable and the description if
+ *  `mergeable` is false
+ */
+module.exports = async (pr, context, settings) => {
+  let body = pr.body
+  let bodySettings = settings.mergeable.body
+  let isMergeable = true
+
+  if (bodySettings === 'non-empty') {
```
jusx: We shouldn't need this conditional. Handler will not call this validator when the user specifies exclusions in the YML:

`excludes: 'body'` 

We don't need settings at all. `Body` validation is still by default validating for zero length body.

lib/body.js
```diff
@@ -0,0 +1,21 @@
...
+    isMergeable = body.length > 0
```
jusx: I think it's a good idea to `trim()` body.

README.md
```diff
@@ -27,6 +27,9 @@ mergeable:
 
   # Minimum of 1 review approval is needed.
   approvals: 1
+  
+  # The body of the PR must not be empty
+  body: 'non-empty'
```
jusx: Should probably remove this now.

shine2lay: removed

PR #18, opened 2018-05-13
enforce consistency in title + tests

### LINKED ISSUE #15: Easily enforce consistency in title
Make it easy to configure mergeable such that the title follows a certain format. For example, in previous projects we've required certain prefixes like `(feat)`, `(doc)`, `(fix)` in the title. This should be easily configured without complex regex like this https://github.com/jusx/mergeable/issues/12#issuecomment-386956281

Instead we should be able to do something like this:

```yml
title: 
   excludes: 'wip'
   includes: '^\(feat\)|^\(doc\)|^\(fix\)' 
```

Probably will still want to be backward compatible with:
```yml
title: 'wip'
```

shine2lay:
closes #15

### DIFF
--- a/__tests__/title.test.js
+++ b/__tests__/title.test.js
@@ -41,7 +41,7 @@ test('checks that it fail when exclude regex is in title', async () => {
 })
 
 test('checks that it fail when include regex is in title', async () => {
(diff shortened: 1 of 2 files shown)


jusx:
Closing this in favor of #19 :-)

PR #19, opened 2018-05-15
enforce consistency in title + tests

### LINKED ISSUE #15: Easily enforce consistency in title

shine2lay:
A second attempt at enforcing consistency in title by putting using a library to allow for re-usability and easily add more filters. 

@jusx I changed the `include` and `exclude` to `must-include` and `must-exclude` respectively to avoid confusion. 
Since we can also add many different filter, i think it make sense to show all the things that failed, I turned the `description` field into an array which holds all the errors. Backward compatibility is maintained so `description` can also be a string as well.

closes #15

### DIFF
--- a/__tests__/title.test.js
+++ b/__tests__/title.test.js
@@ -2,6 +2,17 @@ const Helper = require('../__fixtures__/helper')
 const title = require('../lib/title')
 const Configuration = require('../lib/configuration')
 
+test('fail gracefully if invalid setting is provided', async () => {
+  // try to pass an array instead
+  let config = new Configuration(`
+    mergeable:
+      title: ['item 1', 'item 2' ]
+  `)
+
+  let titleValidation = await title(createMockPR('Valid Title'), null, config.settings)
+  expect(titleValidation.mergeable).toBe(true)
+})
+
 test('fail gracefully if invalid regex', async () => {
   let config = new Configuration(`
     mergeable:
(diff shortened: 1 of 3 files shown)


__tests__/title.test.js
```diff
@@ -2,6 +2,17 @@ const Helper = require('../__fixtures__/helper')
 const title = require('../lib/title')
 const Configuration = require('../lib/configuration')
 
+test('fail gracefully if invalid setting is provided', async () => {
+  // try to pass an array instead
+  let config = new Configuration(`
+    mergeable:
+      title: ['item 1', 'item 2' ]
```
jusx: Interesting. Didn't know you can specify array in YML

lib/processor.js
```diff
@@ -0,0 +1,60 @@
+module.exports = {
+  /**
+   *
+   * @param input the string run regex against
+   * @param filters string | object , if string we check if that regex is excluded from input
+   * @returns object { mergeable: bool, description: string | [string]}
+   */
+  processFilters: function (input, filters) {
```
jusx: We've been using es6 style for functions. Would be good to keep that standard throughout. 

`processFilters: () => {}`

Is it possible to configure standard to error on this I wonder?

lib/processor.js
```diff
@@ -0,0 +1,60 @@
...
+    const isMergeable = regex.test(input)
+    return { mergeable: isMergeable,
+      description: isMergeable ? null : `Title does not contain "${match}"`
+    }
+  },
+  mustExclude: function (input, match) {
+    const regex = new RegExp(match, 'i')
+    const isMergeable = !regex.test(input)
+
+    return { mergeable: isMergeable,
```
jusx: May want to add carriage return here for consistency throughout code base :-)

shine2lay: done

lib/processor.js
shine2lay: done, changed it to `const processFilters = () => {}`

PR #23, opened 2018-05-16
allow advance settings in title

shine2lay:
fix `must-include` to `must_include` instead
added other filter cases: `min`, `max`, `begins_with`, `ends_with`, `no_empty`
allow for sub filter: `regex`, `min`, `max`, `match`, `message` and `enabled`

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -136,8 +136,7 @@ const max = (input, filters) => {
 const beginsWith = (input, filters) => {
   let {match, description} = extractFilterParameter(filters)
   if (!description) description = `Must begins with "${match}"`
-  const regex = new RegExp(`^${match}`, 'i')
-  const isMergeable = regex.test(input)
+  const isMergeable = input.indexOf(match) === 0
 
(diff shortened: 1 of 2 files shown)


lib/processor.js
```diff
@@ -42,23 +59,110 @@ const mergeResults = (result) => {
...
+
+  return {
+    mergeable: isMergeable,
+    description: isMergeable ? null : description
+  }
+}
+
+const max = (input, filters) => {
+  let {max, description} = extractFilterParameter(filters)
+  if (!description) description = `Assignee count is more than "${max}"`
```
jusx: I think that the processor should know who the validator ( validator context). In which case the word `Asignee` would be parameterized. Not a big deal since this is currently for title only -- we can refactor when we propagate these across the other validators.

shine2lay: I think we can add this when we start working on other validator to use this

shine2lay:
@jusx i think this is ready to be merged

PR #42, opened 2018-05-19
fix bug where old approvals are counted toward requirement

### LINKED ISSUE #32: BUG: Approvals count is wrong when reviewer approves and changes mind later
This is primarily an edge case:

## To Reproduce:
- By default the minimum approvals is 1.
- Approve a PR.
- Then change the review status to Comment, which removes your approval status.

## Actual:
Mergeable is true.

## Expected:
Mergeable should be false since the number of approval count is actually zero.

shine2lay:
closes #32 

NOTE: this PR adds a new package `lodash`, npm install will be required after merging

### DIFF
--- a/lib/approvals.js
+++ b/lib/approvals.js
@@ -1,4 +1,5 @@
 const processor = require('../lib/processor')
+const _ = require('lodash')
 /**
  * Determines if the the PR is mergeable based on the number of minimum
  * review approval required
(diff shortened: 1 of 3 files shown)


lib/approvals.js
```diff
@@ -23,3 +26,11 @@ module.exports = async (pr, context, settings) => {
 
   return processor.processFilters('Approval', approvedReviewers, approvalSettings)
 }
+
+const filterOutOldReviews = (reviews) => {
+  const ordered = _.orderBy(reviews, ['submitted_at'], ['desc'])
+  console.log('ordered', ordered)
```
jusx: should remove this.

Issue #40, opened 2018-05-18
Notify Against Consistency in Issues


jusx:
## Context

A standard have been specified for consistencies for Issues.

## Solution
Validate Issues for title, labels, milestones, projects, assignee.

Create a message upon issue.open event.

shine2lay:
I think there is an issue, because when you first open, there will definitely be no label, no milestone, no projects added for sure. This meant that there will be failure. Do you envision this as a reminder?

jusx:
@shine2lay I think you're thinking that this is for PR. This issue is about `issues` though I think we should do the same for `pulls`. In `issues` opened event we should create a comment that displays a "tip" telling the author what he/she missed based on rule-set.

The same should apply for `pulls` -- the intention is that if a team or project sets a particular standards then contributors/members sometimes needs to be reminded of them. Yes, I envision this as a reminder.

So, when someone creates a new issue or pull they don't forget to label, etc.

Longer term vision we want to automate certain things, for example auto-labeling new `issues` as `triage-needed`

PR #48, opened 2018-05-24
(fix) fix tests and libs to support `pull_requests` settings

shine2lay:
These are changes to be able to support #40

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -14,10 +14,11 @@ class Configuration {
 
   loadDefaults () {
     if (this.settings.mergeable == null) this.settings.mergeable = {}
+    if (this.settings.mergeable.pull_requests == null) this.settings.mergeable.pull_requests = {}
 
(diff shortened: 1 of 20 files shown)


lib/configuration.js
```diff
@@ -14,10 +14,11 @@ class Configuration {
 
   loadDefaults () {
     if (this.settings.mergeable == null) this.settings.mergeable = {}
+    if (this.settings.mergeable.pull_requests == null) this.settings.mergeable.pull_requests = {}
 
     for (let key in Configuration.DEFAULTS) {
-      if (this.settings.mergeable[key] === undefined) {
-        this.settings.mergeable[key] = Configuration.DEFAULTS[key]
+      if (this.settings.mergeable.pull_requests[key] === undefined) {
+        this.settings.mergeable.pull_requests[key] = Configuration.DEFAULTS[key]
```
jusx: @shine2lay need to be backward compatible.

PR #50, opened 2018-05-25
remove default approvals count 1

shine2lay:
based on feed backs (some complaint about having default approval count 1), it is removed in this PR

### DIFF
--- a/lib/approvals.js
+++ b/lib/approvals.js
@@ -9,7 +9,7 @@ const _ = require('lodash')
  *  `mergeable` is false
  */
 module.exports = async (pr, context, settings) => {
-  let approvalSettings = settings.mergeable.approvals ? settings.mergeable.approvals : 0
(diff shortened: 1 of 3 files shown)


jusx:
@shine2lay we need to write unit test that if approvals is not defined that the validator does not crash.

shine2lay:
@jusx test added

lib/approvals.js
```diff
@@ -9,7 +9,7 @@ const _ = require('lodash')
  *  `mergeable` is false
  */
 module.exports = async (pr, context, settings) => {
-  let approvalSettings = settings.mergeable.approvals
+  let approvalSettings = settings.mergeable.approvals ? settings.mergeable.approvals : 0
```
jusx: Just a suggestion. But I think this can just be written as `let approvalSettings = settings.mergeable.approvals || 0`

shine2lay: you're right

PR #55, opened 2018-06-06
Added the missing part about local deployments

udaiveerS:
I have changed the order of permissions to the order they appear on GitHub so the users don't have to jump around. I have also added a settings section for the required parameters when setting up a new app. Finally, I have added the steps to run a local deployment.

Let me know what you guys think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,14 +156,20 @@ However you may want to have more advanced rulesets:
 
 ### Deploy your own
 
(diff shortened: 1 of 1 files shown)


udaiveerS:
Yea I will test that

jusx:
@shine2lay @udaiveerS I'm OK to merge this as is. The additional testing and updating of the rest of the README can be in a separate PR.

shine2lay:
@jusx SGTM

PR #57, opened 2018-06-27
Stale Notifications For Issues and Pulls

### LINKED ISSUE #39: Warn Against Stale PRs
## Context

In large repos (usually a monolith) with many contributors the number of PRs can be long. Some PRs become stale and eventually is no longer up to date. 

## Solution
Prompt the author when a PR stays open for `n` number of days.

Configuration:

```
workflow:
   stale:
      pull_request: 
         days: 20 # number of days.
         message: 'some message' # custom message. 
```

jusx:
## Goals
closes #39 

## Details
- Refactor to DRY up helper method for setting context up with a mock config;
- Create comment on issues and/or Pulls that are greater than `n` number of days.
- Default message for stale warning 
- Full unit test coverage for stale.
- Manually tested with sandbox.

Configuration is as follows:
```yml
issues:
   stale:
      days: 10
pull_requests:
   stale:
      days: 5
```
If `message` is not set, then the default is used.  Both are optional. When `stale.days` properties for issues or pulls are configured then stale warnings will be created as a comment.


jusx:
Putting this on WIP because I still need to update README.

lib/stale.js
```diff
@@ -0,0 +1,48 @@
...
+ * is set for Pulls and Issues respectively.
+ *
+ * Assumes that the numbers of days are always set if stale property in Settings
+ * is defined.
+ *
+ * @return
+ *  JSON object with the properties mergeable and the description if
+ *  `mergeable` is false
+ */
+module.exports = async (context, settings) => {
```
shine2lay: this function seems redundant, i think you can just do `module.exports = process`

lib/stale.js
```diff
@@ -0,0 +1,48 @@
...
+  return process(context, settings)
+}
+
+const process = async (context, config) => {
```
shine2lay: i notice that these functions doesn't return anything, i am curious about what the behavior would be if we `await process()`.

lib/configuration.js
```diff
@@ -22,6 +22,18 @@ class Configuration {
         this.settings.mergeable[key] = Configuration.DEFAULTS[key]
       }
     }
+
+    // ensure stale defaults
+    if (this.settings.mergeable.pull_requests &&
+      this.settings.mergeable.pull_requests.stale !== undefined &&
```
shine2lay: i think we don't need `!== undefined` here. because we also don't want the `stale` to be null right?

index.js
```diff
@@ -11,14 +12,20 @@ module.exports = (robot) => {
...
-    (context) => { Handler.handleIssuesOpened(context) }
+    context => Handler.handleIssues(context)
+  )
+
+  robot.on(['issues.opened'],
+    (context) => { Handler.handleIssuesOpened(context) }
+  )
+
+  // check every two seconds.
+  scheduler(robot, { interval: 60 * 60 * 2 })
```
shine2lay: i am not sure how scheduler works but from the comment it seems like we check for staleness every two seconds is that correct? if so is there any check to make sure we are not spamming `createComment` every two seconds?

docs/README.md
```diff
@@ -221,7 +234,11 @@ Here's an example configuration file for advanced settings and all of it's possi
     #  Advanced settings for issues. When any of the rules  below is not valid Mergeable will create a comment on that issue to let the author know.    
     ###
     issues:
-
+      stale:
+        # number of days for an issue to be considered stale.  
+        days: 20
+        # Optional property. When not specified the default is used. The default message is used.
+        message: 'This is issue is stale. Please followup!'
```
shine2lay: s/followup/follow up

docs/README.md
```diff
@@ -136,7 +144,12 @@ Here's an example configuration file for advanced settings and all of it's possi
   ##############################
   mergeable:
     pull_requests:
-
+      stale:
+        # number of days for a PR to be considered stale.  
+        days: 20
+        # Optional property. When not specified the default is used. The default message is:
+        # There haven't been much activity here. This is stale. Is it still relevant? This is a friendly reminder to please resolve it. :-)
+        message: 'This is PR is stale. Please followup!'
```
shine2lay: s/followup/follow up

__tests__/stale.test.js
```diff
@@ -0,0 +1,79 @@
...
+    mergeable:
+      issues:
+        stale:
+          days: 20
+    `)
+
+  return context
+}
+
+const createMockContectWithPullsSetting = (results) => {
```
shine2lay: s/MockContect/MockContext

__tests__/stale.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+test('will NOT create comment when configured and stale issues are not found.', async () => {
+  let context = createMockContectWithPullsSetting([])
+  let config = await Configuration.instanceWithContext(context)
+
+  await stale(context, config)
+  expect(context.github.issues.createComment.mock.calls.length).toBe(0)
+})
+
+const createMockContectWithIssueSetting = (results) => {
```
shine2lay: s/MockContect/MockContext

lib/stale.js
jusx: It is not redundant. I have to return the promise so that if it is rejected jest will fail the test.

lib/configuration.js
jusx: well, currently `stale` is undefined if it is not specified in the yml.

index.js
jusx: `createComment` will cause the issue/pull to have an updated timestamp and hence the search query will not return that issue/pull again till there's no activity since `n` days.

lib/configuration.js
jusx: I see what you're saying. You're right.

PR #65, opened 2018-07-05
(feat) check files for specific header

### LINKED ISSUE #53: Check copyright and license notices
Some projects require that every source file starts with a piece of text related to license and copyright.

The CI could block a PR if some new files are introduced without these lines.

shine2lay:
Goals
---
closes #53 

Details 
---
- Added a new validation to check file header for specific string (i.e license) 
- fix a bug in parseOwner where `test.js` fails to match but `**/test.js` passes


docs/README.md
```diff
@@ -224,6 +224,14 @@ Here's an example configuration file for advanced settings and all of it's possi
...
           regex: 'DO NOT MERGE'
           message: 'Custom message...'
+          
+      files:
+		  # Check each modified file in the PR for the following starting at line 0.
+		  header: > 
+			   Entire license header.
+		  # only enforce it for these file patterns uses the same rules as CODEOWNER file
+		  pattern: *.go
+		  message: 'Custom message...'
```
jusx: Looks like some kind of tabbing issue? Can you fix?

lib/dependent.js
```diff
@@ -22,6 +22,7 @@ module.exports = async (pr, context, settings) => {
 
   // fetch the file list
   let result = await context.github.pullRequests.getFiles(context.repo({number: pr.number}))
+
   let modifiedFiles = result.data.filter(file => file.status === 'modified').map(file => file.filename)
```
jusx: Looks like an accident CR here? :-)

lib/parseOwners.js
```diff
@@ -118,4 +123,7 @@ const arrayify = (val) => {
   return Array.isArray(val) ? val : [val]
 }
 
-module.exports = parseOwners
+module.exports = {
+  parseOwners: parseOwners,
+  checkIfPathMatches: checkIfPathMatches
```
jusx: IMO It will be cleaner to extract this out particular generic function out to a "FileUtil" module. As it is now used in two places. Here and outside of this with a totally responsibility of work than parseOwners.

lib/files.js
```diff
@@ -0,0 +1,54 @@
+const { checkIfPathMatches } = require('./parseOwners')
```
jusx: This is very cool. I had no idea you can do this with require. :-)

lib/files.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * process the PR based on dependent settings
```
jusx: Probably want to update the description here.

lib/files.js
```diff
@@ -0,0 +1,54 @@
...
+  let addedFiles = result.data.filter(file => file.status === 'added')
+  // parse the files to match the pattern
+  const filesToCheck = addedFiles.filter(file => checkIfPathMatches(file.filename, patternToMatch))
+
+  // fetch the file content
+  result = await fetchFileContents(filesToCheck, context)
+  const failedFiles = result.filter(file => {
+    const content = Buffer.from(file.data.content, 'base64').toString()
+
+    return !content.includes(settings.files.header)
```
jusx: If I'm not mistaken this only checks if the content exists in the string and not necessarily from line position 0? Perhaps `startsWith` will be more appropriate here?

jusx: Also probably want to add a test where "Test Header" is in the middle of the content of a file. And it should return false...

lib/files.js
```diff
@@ -0,0 +1,54 @@
...
+    return {
+      mergeable: true,
+      description: 'Okay To Merge'
+    }
+  }
+
+  let patternToMatch = settings.files.pattern
+  // fetch the file list
+  let result = await context.github.pullRequests.getFiles(context.repo({number: pr.number}))
+  let addedFiles = result.data.filter(file => file.status === 'added')
```
jusx: Any reason why you aren't including `modified` in here?

docs/README.md
shine2lay: done

lib/dependent.js
shine2lay: done

lib/files.js
shine2lay: :+1:

lib/files.js
shine2lay: done

lib/files.js
shine2lay: no reason, fixed

lib/parseOwners.js
shine2lay: this file does everything that has to do with parsing gitpattern, i don't think we need to refactor it out to a `fileUtils`, i'll rename the file to `gitpatternParser`

lib/files.js
shine2lay: done

docs/README.md
```diff
@@ -229,6 +229,14 @@ Here's an example configuration file for advanced settings and all of it's possi
         must_exclude:
           regex: 'DO NOT MERGE'
           message: 'Custom message...'
+          
+      files:
+        # Check each modified file in the PR for the following starting at line 0.
+        header: > 
+          Entire license header.
+        # only enforce it for these file patterns uses the same rules as CODEOWNER file
```
jusx: The `CODEOWNERS` file follows the same file matching patterns as the `gitignore`. I suggest having this link (https://git-scm.com/docs/gitignore#_pattern_format) in the comment above to be clear.

lib/files.js
```diff
@@ -0,0 +1,54 @@
...
+
+  let patternToMatch = settings.files.pattern
+  // fetch the file list
+  let result = await context.github.pullRequests.getFiles(context.repo({number: pr.number}))
+  let addedFiles = result.data.filter(file => file.status === 'added' || file.status === 'modified')
+  // parse the files to match the pattern
+  const filesToCheck = addedFiles.filter(file => checkIfPathMatches(file.filename, patternToMatch))
+
+  // fetch the file content
+  result = await fetchFileContents(filesToCheck, context)
```
jusx: Might be cleaner to write the code block above (from L21 to L27) as so:
```js
let filesToCheck = await context.github.pullRequests.getFiles(context.repo({number: pr.number})).data
   .filter(file => file.status === 'added' || file.status === 'modified')
   .filter(file => checkIfPathMatches(file.filename, patternToMatch))

let fileContents = await fetchFileContents(filesToCheck, context)
```

Here's why I think so:
- The variable `result` is quickly overwritten with some other function call result and is too loosely defined. Not obviously apparent at first glance.
- `addedFiles` is very temporary and is not used elsewhere except very subsequently to filter again. It's better not to declare it. The reader will not wonder what else it is used for.

lib/files.js
```diff
@@ -0,0 +1,54 @@
...
+  const failedFiles = result.filter(file => {
```
jusx: I noticed you've been using `const` and `let` interchangeably. I'm wondering what the reasoning behind it is?

So far we've primarily used `let` for variables and `const` for imported modules and function.

lib/files.js
```diff
@@ -0,0 +1,54 @@
...
+    const content = Buffer.from(file.data.content, 'base64').toString()
+
+    return !(content.indexOf(settings.files.header) === 0) // first instead of header must begin at position 0 to be a header
+  }).map(file => file.data.path)
+
+  // check the header of each file
+
+  const isMergeable = failedFiles.length === 0
```
jusx: IMO this expression is short enough to be in line 42. Hence you won't need this additional line.

```
mergeable: (failedFiles.length === 0)
```

lib/parseOwners.js
jusx: fair enough.

lib/files.js
shine2lay: done

lib/files.js
shine2lay: i prefert `const` over `let` because we prevent accidental overwriting of variables. It is just safer in general IMHO, the reason i sometime end up using let is because of my habit. thanks for pointing it out

lib/files.js
shine2lay: `isMergeable` is used twice, and i like having this variable because it makes it much clear to me what the mergeable conditional are

docs/README.md
shine2lay: done

shine2lay:
a

shine2lay:
based on our discussion offline, we will not support this feature until future versions

PR #66, opened 2018-07-06
Bugfix: Stale Reminder Performance

jusx:
We are scheduling checks to fetch every 2 seconds. That's too much. It will now just use the default of every 1 hour.

Every 2 secs will cause performance problems on first install into an org with a lot of repos.


jusx:
@shine2lay FYI incremented the milestone in mergeable yml.

PR #73, opened 2018-07-27
show `action_requested` if error occurs

### LINKED ISSUE #72: YML Exception Should Bubble Up
### Context
When the mergeable YML does not have expected indentation or the format is wrong `js-yaml` throws an exception and mergeable does not stop running.

### Proposal
- Mergeable should catch these exceptions and notify.
- **Potential Notifications:**
   - Set a status on PR with bad YML configuration found. Ideally the details should link to a mergeable
   - Send a checks with explanation about YML problem.
   - Comment on the PR or issue when the problem is found.

shine2lay:
closes #72 

Screen shot of the error message
image

### DIFF
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -113,9 +113,6 @@ class Handler {
             summary: description})
 
         console.info({state: status, description: description})
-      }).catch(error => {
-        // (jusx) This should never ever happen. Log it.
-        console.error(error)
       })
     } catch (err) {
       checks.update(
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -73,47 +73,9 @@ class Handler {
   static async handle (context, pullRequest) {
     const checkRunResult = await checks.create(context, 'Mergeable')
     // let the user know that we are validating if PR is mergeable
+    let config
     try {
-      var config = await Configuration.instanceWithContext(context)
-      let settings = config.settings.mergeable.pull_requests || config.settings.mergeable
-      let validators = []
-      let excludes = (settings.exclude)
-        ? settings.exclude.split(',').map(val => val.trim()) : []
-      let includes = [ 'approvals', 'title', 'label', 'milestone', 'description', 'projects', 'assignee', 'dependent' ]
-        .filter(validator => excludes.indexOf(validator) === -1)
-
-      includes.forEach(validator => {
-        validators.push(require(`../lib/${validator}`)(pullRequest, context, settings))
-      })
-      console.info(config)
-      return Promise.all(validators).then(results => {
-        let failures = results.filter(validated => !validated.mergeable)
-
-        let status, description
-        if (failures.length === 0) {
-          status = 'success'
-          description = 'Okay to merge.'
-        } else {
-          status = 'failure'
-          description = `## Mergeable has found the following failed checks`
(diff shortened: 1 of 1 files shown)


lib/handler.js
```diff
@@ -73,52 +73,61 @@ class Handler {
...
-      if (failures.length === 0) {
-        status = 'success'
-        description = 'Okay to merge.'
-      } else {
-        status = 'failure'
-        description = `## Mergeable has found the following failed checks`
-        for (let validated of failures) {
-          description += `\n - ${validated.description}`
+    try {
+      var config = await Configuration.instanceWithContext(context)
```
jusx: IMO the try catch should be around this statement where it is an actual error from loading / processing the configuration.

lib/handler.js
```diff
@@ -73,52 +73,61 @@ class Handler {
...
+          'completed',
+          status,
+          {
+            title: `Result: ${status}`,
+            summary: description})
+
+        console.info({state: status, description: description})
+      }).catch(error => {
+        // (jusx) This should never ever happen. Log it.
+        console.error(error)
```
jusx: There should probably be a `checks.update` here for general errors. Though likely no action required because it isn't really configuration related. This will likely be like a "HTTP 500".

lib/handler.js
shine2lay: i thought we want to do a general try and catch for all unknown errors

lib/handler.js
shine2lay: removed that `catch` the try and catch block should catch this error as well

lib/handler.js
```diff
@@ -73,52 +73,58 @@ class Handler {
...
         'Mergeable',
         'completed',
-        status,
+        'action_required',
         {
-          title: `Result: ${status}`,
-          summary: description})
-
-      console.info({state: status, description: description})
-    }).catch(error => {
```
jusx: This catches `promise` rejections. Removing this will cause unknown errors to go unlogged. I really don't think it should be removed. Normal try catches will do anything here as this is async.

Furthermore at this point it's errors thrown by validators are not due to bad configuration and are typically errors caused by system bugs snda hence the message should tell the user that a "system error has occurred"  -- it's not something they can do on their end. I also really would like these errors to be logged.

lib/handler.js
jusx: The original catch was on `Promise.all` and would have caught any promise rejections in this entire function.

The only other place where an error could occur is during the loading of Configuration. A try catch around that with the very specific message telling the user to fix their configuration will suffice. It is actionable. The promise rejections from the validators are unlikely to be actionable. It deserves it's own message and an error log since it's most likely to be a bug.

lib/handler.js
```diff
@@ -115,9 +126,16 @@ class Handler {
           summary: description})
 
       console.info({state: status, description: description})
-    }).catch(error => {
-      // (jusx) This should never ever happen. Log it.
-      console.error(error)
+    }).catch(err => {
+      checks.update(
```
jusx: We should log the error as well so that we are aware of it and can fix the bug.

shine2lay: done

PR #86, opened 2018-08-19
owners option + gitPattern, and tests

shine2lay:
@jusx This is also another independent, ready to be merged


jusx:
I will wait to review this until we resolve comments in #85

PR #88, opened 2018-08-19
action class + test

shine2lay:
action class

### DIFF
--- a/__tests__/actions/action.test.js
+++ b/__tests__/actions/action.test.js
@@ -1,6 +1,6 @@
 const { Action } = require('../../lib/actions/action')
 
-describe('Validator#isEventSupported', () => {
+describe('Action#isEventSupported', () => {
   let action = new Action()
 
   test('Returns correctly with one supported event', () => {
--- a/__tests__/actions/action.test.js
+++ b/__tests__/actions/action.test.js
@@ -1,0 +1,24 @@
+const { Action } = require('../../lib/actions/action')
+
+describe('Validator#isEventSupported', () => {
+  let action = new Action()
+
+  test('Returns correctly with one supported event', () => {
+    action.supportedEvents = ['issues.opened']
+    expect(action.isEventSupported('pull_request.opened')).toBe(false)
+    expect(action.isEventSupported('issues.opened')).toBe(true)
+  })
+
+  test('Returns correctly with more than one supported event', () => {
+    action.supportedEvents = ['issues.opened', 'pull_request.opened']
+    expect(action.isEventSupported('pull_request.labeled')).toBe(false)
+    expect(action.isEventSupported('issues.opened')).toBe(true)
+    expect(action.isEventSupported('pull_request.opened')).toBe(true)
+  })
+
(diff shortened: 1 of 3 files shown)


__tests__/actions/action.test.js
```diff
@@ -0,0 +1,24 @@
+const { Action } = require('../../lib/actions/action')
+
+describe('Validator#isEventSupported', () => {
```
jusx: Should be `#ActionisEventSupported`

lib/actions/action.js
```diff
@@ -0,0 +1,28 @@
+class Action {
```
jusx: This class shares a lot of the same base traits as the Validator base class. I propose:

- Both Validator and Action extends `EventAware` which includes all relevant shared functions.
- `EventAware` can reside in `lib/`

lib/actions/action.js
```diff
@@ -0,0 +1,28 @@
...
+   * @param eventName
+   *  An event name to be evaluated for support. The name is as in the GitHub
+   *  webhook format of issues.opened, pull_request.opened, etc
+   *
+   * @return boolean true if the validator supports the event. i.e. issues.opened
+   */
+  isEventSupported (eventName) {
+    return isEventInContext(eventName, this.supportedEvents)
+  }
+
```
jusx: I recommend declaring an `async` method here called `execute` -- which is the common interface for all actions to be executed by the `Mergeable` / executor. In this implementation it would throw an error. Here's why:

There's are no interfaces in JS but the caller needs a common interface to call, in this case it's `execute` -- if for any reason a child class forgot to override this method correctly then at least the system can return appropriate errors as opposed to unknown errors.

__tests__/actions/action.test.js
shine2lay: done

lib/actions/action.js
shine2lay: i have a different methods for action than validators, i have `doPreAction` and `doPostAction`, this is for actions like checks which may require we do something before we run validation, idk how execute will fit into this

lib/actions/action.js
shine2lay: done

lib/actions/action.js
jusx: Based on the names of those methods. `doPreAction` is called before the "action" method is invoked and `doPostAction` is called after the "action" method is invoked -- Is that correct? What is the method that is in between the two? Or is there none?

shine2lay: hmm.. i think the better names would be `doPreValidationAction`, `doPostValidationAction`. for example, in checks, preValidationAction would be to create a pending status, post would update it to completed status

jusx: Action is something that executes after validation. The main method for execution is after validation is done. I fine it totally OK to name it `execute` and having a `beforeValidation` method.

Alternatively If you want to be more explicit which might be better -- you can name them `beforeValidation` and `afterValidation`. Both are familiar naming conventions in frameworks.

jusx: I do recommend to declare them in this base class and have them throw Error.

PR #95, opened 2018-08-25
stale validation + test

shine2lay:
stale

### DIFF
--- a/lib/validators/stale.js
+++ b/lib/validators/stale.js
@@ -4,15 +4,7 @@ class Stale extends Validator {
   constructor () {
     super()
     this.supportedEvents = [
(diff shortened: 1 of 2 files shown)


lib/validators/stale.js
```diff
@@ -0,0 +1,52 @@
+const { Validator } = require('./validator')
+
+class Stale extends Validator {
+  constructor () {
+    super()
+    this.supportedEvents = [
```
jusx: The event here is wrong. It should be `schedule` -- and just that. Have you tested it with these events?

shine2lay: that is true, i'll change it

PR #96, opened 2018-09-07
Flex class

shine2lay:
Main logic class


lib/flex.js
```diff
@@ -0,0 +1,132 @@
...
+    let events = Array.isArray(rule.when) ? rule.when : rule.when.split(',').map(n => n.trim())
+    if (isEventInContext(eventName, events)) {
+      filteredConfig.push(rule)
+    }
+  })
+
+  return filteredConfig
+}
+
+const isEventInContext = (eventName, events) => {
```
jusx: This function is not used anywhere. What is it for?

lib/flex.js
```diff
@@ -0,0 +1,132 @@
...
+    (events.indexOf('issues.*') > -1 && eventObject === 'issues'))
+}
+
+const registerValidatorsAndActions = (config, registry) => {
+  config.forEach(rule => {
+    rule.validate.forEach(validation => {
+      let key = validation.do
+      if (!registry.validators.has(key)) {
+        if (!SUPPORTED_VALIDATOR.includes(key)) {
+          return // add validator not supported error
```
jusx: This should throw an Error.

jusx: Also shouldn't this be validated in `Configuration`?

lib/flex.js
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        let Action = require(`./actions/${key}`)
+        registry.actions.set(key, new Action())
+      }
+    })
+  })
+}
+
+const getValiationPromises = (context, registry, rule) => {
```
jusx: Should be called `getValidatorPormises` IMO since our classes are named `Validators`

lib/flex.js
```diff
@@ -0,0 +1,132 @@
...
+    if (validator.isEventSupported(eventName)) {
+      promises.push(validator.validate(context, validation))
+    } // else add error for event not supported
+  })
+
+  return promises
+}
+
+// call all action classes' preAction, regardless of whether they are in failure or pass situation
+const processPreActions = async (context, registry, config) => {
```
jusx: Since the interface in an `Action` is called `beforeValidate` why not call this method `processBeforeValidate` to be consistent with terminology?

lib/flex.js
```diff
@@ -0,0 +1,132 @@
...
+  registry.actions.forEach(action => {
+    action.beforeValidate({ context })
+  })
+}
+
+const getActionPromises = (context, registry, rule, result) => {
```
jusx: There is a lot of code here that is exactly like `getValidatorPromises` An opportunity to generalize this function. A refactor for the future perhaps?

lib/flex.js
```diff
@@ -0,0 +1,132 @@
...
+  'title',
+  'label',
+  'description',
+  'assignee',
+  'dependent',
+  'milestone',
+  'approval'
+]
+
+const SUPPORTED_ACTIONS = [
```
jusx: In the long run I would rather we read this from a path thst can be overriden but I think these constants work for now.

jusx: Also we don't really need these arrays. You could just catch the error of the require statement and take appropriate actions.

lib/flex.js
```diff
@@ -0,0 +1,132 @@
...
+  'comments',
+  'checks'
+]
+
+// Main logic Processor of mergeable
+const executeMergeable = async (context, registry) => {
```
jusx: Where is this being called? And how? Is `/index.js` missing from this PR? Is `lib/mergeable.js` missing from this PR?

lib/flex.js
shine2lay: it is being used in `filterByEvent` but i agree that configuration should be handling this code

lib/flex.js
shine2lay: yeah, try and catch would be better

lib/flex.js
shine2lay: done

lib/flex.js
shine2lay: done

lib/flex.js
shine2lay: done

lib/flex.js
shine2lay: done

lib/flex.js
shine2lay: done

lib/flex.js
shine2lay: done

lib/flex.js
```diff
@@ -0,0 +1,69 @@
...
+  if (registry === undefined) {
+    registry = { validators: new Map(), actions: new Map() }
+  }
+
+  // first fetch the configuration
+  let config = await Configuration.instanceWithContext(context)
+
+  // check config file and get settings only for the current event. the rest is useless for executor
+  let eventName = `${context.event}.${context.payload.action}`
+  await config.filterByEvent(eventName)
```
jusx: Any reason for filtering events? I propose we don't filter. Here's why:
- If there are no validators that supports an event, the rule will just be ignored. 
- Adding new validators or updating a validator to support new events will require update to the Configuration class filtering logic as well. Coupling all validators to Configuration.

lib/configuration/configuration.js
```diff
@@ -102,6 +144,13 @@ class Configuration {
   }
 }
 
+const isEventInContext = (eventName, events) => {
+  let eventObject = eventName.split('.')[0]
+  return ((events.indexOf(eventName.trim()) > -1) ||
+    (events.indexOf('pull_request.*') > -1 && eventObject === 'pull_request') ||
+    (events.indexOf('issues.*') > -1 && eventObject === 'issues'))
```
jusx: missing `schedule.*`

shine2lay: i generalized by using `${eventObject}.* instead, this should cover all cases

lib/flex.js
```diff
@@ -0,0 +1,66 @@
...
+  return createPromises(actions, 'actions', afterValidateFuncCall, context, registry, result)
+}
+
+const createPromises = (arrayToIterate, registryName, funcCall, context, registry, result) => {
+  let promises = []
+
+  arrayToIterate.forEach(element => {
+    let key = element.do
+
+    let actionClass = registry[registryName].get(key)
```
jusx: I think this variable name should probably just be `klass`  since it can be any type of class. and `class` is a reserved word.

jusx: Or something else more general.

lib/eventAware.js
```diff
@@ -18,8 +18,7 @@ class EventAware {
 const isEventInContext = (eventName, events) => {
   let eventObject = eventName.split('.')[0]
   return ((events.indexOf(eventName.trim()) > -1) ||
-    (events.indexOf('pull_request.*') > -1 && eventObject === 'pull_request') ||
-    (events.indexOf('issues.*') > -1 && eventObject === 'issues'))
+    (events.indexOf(`${eventObject}.*`) > -1))
```
jusx: Should be:

```js
events.map(event=>event.split('.')[0]).indexOf(eventObject)
```

lib/flex.js
shine2lay: `klass` is cool, sold

lib/eventAware.js
```diff
@@ -17,9 +17,14 @@ class EventAware {
...
   let eventObject = eventName.split('.')[0]
-  return ((events.indexOf(eventName.trim()) > -1) ||
-    (events.indexOf('pull_request.*') > -1 && eventObject === 'pull_request') ||
-    (events.indexOf('issues.*') > -1 && eventObject === 'issues'))
+  let relevantEvent = []
+  events.forEach(event => {
+    if (event.split('.')[0] === eventObject) {
+      relevantEvent.push(event)
+    }
+  })
```
jusx: not a big deal, but this can be written succinctly with 
```js
let relevantEvents = events.map( event => event.split('.')[0] === eventObject )
```

PR #100, opened 2018-10-02
fix #98 + test to mimic the same scenario

### LINKED ISSUE #98: Title: must_include: regex fails with "TypeError: Cannot read property 'trim' of null"
When trying to validate pull request titles, mergeable fails with the following:

image

I've tried using the example from the readme:

```
title:
  must_include:
    regex: `^\\(feat\\)|^\\(doc\\)|^\\(fix\\)`
      message: `Title must have prefixes for the following: (feat), (doc), (fix)`
```

and the following:

```
  must_include:
    regex: `#\d{7,}'
```

shine2lay:
closes #98 
Please see comments for more information about the bug https://github.com/jusx/mergeable/issues/98#issuecomment-426302899

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -6,47 +6,48 @@
  */
 const processFilters = (validatorContext, input, filters) => {
   // if input is null or undefined, turn it into empty string
-  if (!input) {
-    input = ''
+  let definedInput = ''
+  if (!!input) {  // eslint-disable-line
+    definedInput = input
   }
 
   // for backward compatibility, check if the filters is a string
   if (typeof filters === 'string') {
-    return mustExclude(validatorContext, input, filters)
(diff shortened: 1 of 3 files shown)


lib/processor.js
```diff
@@ -5,6 +5,11 @@
  * @returns object { mergeable: bool, description: string | [string]}
  */
 const processFilters = (validatorContext, input, filters) => {
+  // if input is null or undefined, turn it into empty string
+  if (!input) {
```
jusx: To be on the safe side I propose using a double bang here. i.e. `(!!input)`

shine2lay: standard said `no double negation`, would you prefer i don't fix this ? or fix standard?

PR #102, opened 2018-10-07
Default Checks output template

shine2lay:
Context 
---
We want to display information of each validations that that were ran and display it in checks by default.

Things Changed
---
- Added validator name field to outputs from processor
- Added handlebars template string to go through each validation suite and display them

Note: Not well tested yet! [EDIT: it's working]

Default template looks like following
---
image
image

### DIFF
--- a/lib/configuration/configuration.js
+++ b/lib/configuration/configuration.js
@@ -59,7 +59,10 @@ class Configuration {
     this.settings.forEach(rule => {
       try {
         rule.validate.forEach(validation => {
+          console.log('validation', validation)
+          console.log('key', validation.do)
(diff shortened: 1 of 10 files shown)


shine2lay:
@jusx I believe this is ready

lib/configuration/configuration.js
```diff
@@ -59,7 +59,10 @@ class Configuration {
     this.settings.forEach(rule => {
       try {
         rule.validate.forEach(validation => {
+          console.log('validation', validation)
+          console.log('key', validation.do)
```
jusx: Did you mean for these console.logs to be here?

PR #103, opened 2018-10-08
Log Mergeable Version During Startup

jusx:
### Goal

To clearly show the version that Mergeable is booted up in.


jusx:
Opened against the wrong branch. Closing.

PR #110, opened 2018-11-25
if event is pull_request, use the config file from head branch of PR

shine2lay:
Goal
---
When an Event is Pull_request, use the head branch's config file instead of the master branches

Context
---
This will allow user to test different settings and see how it works before merging it into the code base

### DIFF
--- a/lib/configuration/configuration.js
+++ b/lib/configuration/configuration.js
@@ -133,7 +133,6 @@ class Configuration {
   }
 
   static instanceWithContext (context) {
-
     return Configuration.fetchConfigFile(context).then(res => {
       let content = Buffer.from(res.data.content, 'base64').toString()
 
--- a/lib/configuration/configuration.js
+++ b/lib/configuration/configuration.js
@@ -112,15 +112,29 @@ class Configuration {
     }
   }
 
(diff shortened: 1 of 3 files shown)


lib/configuration/configuration.js
```diff
@@ -112,15 +112,29 @@ class Configuration {
...
-      repo: repo.repo,
-      path: Configuration.FILE_NAME
-    }).then(res => {
+    if (context.event === 'pull_request') {
+      return github.repos.getContent({
+        owner: repo.owner,
+        repo: repo.repo,
+        path: Configuration.FILE_NAME,
+        ref: context.payload.pull_request.head.ref
+      })
```
jusx: What happens here when there is no Configuration changes or additions in the PR? I think in that scenario it still needs to retrieve from the repo "master"

jusx:
@shine2lay We should probably also update the unit tests.

shine2lay:
@jusx updated

lib/configuration/configuration.js
jusx: OK sounds good. If that's how the `github.repos.getContent` we are gtg.

PR #119, opened 2018-12-20
Remove redundant console logs

jusx:
### Goal
- Remove redundant console logs


shine2lay:
@jusx It said there are no diffs in your PR

PR #120, opened 2018-12-20
Feature: Default Actions for V2 Config

jusx:
### Goal
- Feature: To ensure that there are default `pass`, `fail`, `error` actions for pull_request recipes.
- Refactor: Rename constant so that it is clear it's for `pull_request` events only.


shine2lay:
@jusx i am going to merge this in, i think this is going to fix something i am working on

jusx:
ok

PR #123, opened 2018-12-22
add assigned and unassigned to support events

shine2lay:
Goal
---
Fix that checks weren't running when assignee is added/removed

### DIFF
--- a/lib/validators/assignee.js
+++ b/lib/validators/assignee.js
@@ -22,7 +22,7 @@ class Description extends Validator {
 
   async validate (context, validationSettings) {
     let assignees = this.getPayload(context).assignees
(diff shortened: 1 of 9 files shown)


lib/validators/assignee.js
```diff
@@ -13,17 +13,19 @@ class Description extends Validator {
...
+      'pull_request.unassigned',
       'pull_request.synchronize',
       'issues.opened'
     ]
   }
 
   async validate (context, validationSettings) {
     let assignees = this.getPayload(context).assignees
-
+    console.log('assignee', assignees.map(assignee => assignee.login))
```
jusx: Not a big deal but we may want to remove this console log.

shine2lay: done

PR #124, opened 2018-12-22
fix milestone deep validation

shine2lay:
Goals
---
fix deep validation for milestone


shine2lay:
this is a very edge case that occured, we have tests that covered the happy path already, test will be added in a different PR

PR #130, opened 2019-01-07
fix: dependent to support added files

jusx:
### Goal
Fixes `dependent` validators' `files` option to also validate against `added` files in a PR.

### Changes
- Updated test helper to support more detailed files with status metadata to be passed in at `options.files`
- Updated dependent unit test
- Fixed validator to include files that are also `added`
- Changed default message to include file that is missing so that it is more explicit and useful.

### DIFF
--- a/__tests__/validators/dependent.test.js
+++ b/__tests__/validators/dependent.test.js
@@ -72,7 +72,7 @@ test('test description is correct', async () => {
     files: ['package.json', 'yarn.lock']
   }
 
-  let defaultMessage = 'One or two files (yarn.lock) is missing from your pull request because they are dependent on the following: package.json'
+  let defaultMessage = 'One or more files (yarn.lock) are missing from your pull request because they are dependent on the following: package.json'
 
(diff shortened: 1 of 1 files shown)


lib/validators/dependent.js
```diff
@@ -28,19 +28,22 @@ class Dependent extends Validator {
...
+      .map(file => file.filename)
     const validatorContext = {name: 'Dependent'}
+
     if (!dependentFiles) {
       return consolidateResult([constructError('Dependent', modifiedFiles, validationSettings, FILES_NOT_FOUND_ERROR)], validatorContext)
     }
     const fileDiff = _.difference(dependentFiles, modifiedFiles)
 
+    let description = validationSettings.message ||
+      `One or two files (${fileDiff.join(', ')}) is missing from your pull request because they are dependent on the following: ${_.difference(dependentFiles, fileDiff)}`
```
shine2lay: I think this default message is misleading, i think it would be better to use `One or more files` instead

jusx: Good catch! I'll update :-)

PR #133, opened 2019-01-09
fix: stale

jusx:
### Goal
Make `stale` work in flex. 

### Changes
- Update comment to execute based on results of validators. Look for schedule results and comment appropriately.
- Update v1 transformer to output correct format for schedule.repository event.
- Update `Stale` validator to focus only on finding stale issues and/or pr -- it shouldn't be creating comments in the new flex.
- Update all relevant unit tests

### DIFF
--- a/__tests__/validators/stale.test.js
+++ b/__tests__/validators/stale.test.js
@@ -34,7 +34,7 @@ test('will set the issues and pulls appropriately when no type is set', async ()
   expect(results.schedule.pulls.length).toBe(1)
 })
 
-test('will set the issues and pulls even when unsupported type is set -- nothing', async () => {
+test('will set the issues and pulls even when unsupported type is set', async () => {
   let settings = { do: 'stale', days: 10, type: ['junk1', 'junk2'] }
 
   let stale = new Stale()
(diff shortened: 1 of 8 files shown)


__tests__/validators/stale.test.js
```diff
@@ -94,6 +96,11 @@ const createMockContext = (results) => {
     })
   }
 
-  context.github.issues.createComment = jest.fn()
+  // context.github.search = {
```
shine2lay: please remove un need codes

lib/configuration/transformers/v1Config.js
```diff
@@ -35,21 +36,37 @@ class V1Config {
...
     if (!config.pull_requests && !config.issues) {
       output.push(processValidators('pull_request.*', config, {pass: consts.DEFAULT_PR_PASS, fail: consts.DEFAULT_PR_FAIL, error: consts.DEFAULT_PR_ERROR}))
       output.push(processValidators('pull_request_review.*', config, {pass: consts.DEFAULT_PR_PASS, fail: consts.DEFAULT_PR_FAIL, error: consts.DEFAULT_PR_ERROR}))
     }
-
     return {mergeable: output}
   }
 }
 
+const processStale = (type, config, output) => {
```
shine2lay: just double checking , did we have `stale` in the first version config where we don't need `issues` or `pull_request`?

__tests__/validators/stale.test.js
jusx: Done! Remnants of debugging code. Thanks for pointing it out. :-|

lib/configuration/transformers/v1Config.js
jusx: No. But the format is different. In v1, everything either fall under `issues` or `pull_requests` at the parent node. So, there really is no way to specify `stale` without needing either `issues` or `pull_requests`.

PR #134, opened 2019-01-10
chore: new readme 

jusx:
### Goal
Update README to reflect new changes in flex.

### Changes
- Entire README
- Logo
- License to AGPL.


docs/README.md
```diff
@@ -19,417 +17,498 @@
...
 
-- Ensure that all Pull Requests are approved by a specific list of users. It is especially useful if one of the users is not a collaborator of your repository -- something GitHub does not already support.
+> ☝ **NOTE:** You can also [deploy to your own server](deploy.md).
 
-- Ensure that all Pull Requests merged are in a specific GitHub Project. Mergeable even detects when you are closing an issue that is associated with the right project. This is very useful when your process includes QA validation of stories.
+# Configuration
 
-- [And more](#configuration).
+**Mergeable** is **highly** configurable.
+Define your recipes by creating a `.github/mergeable.yml` file in your repository.
```
shine2lay: Would it make more sense to say `Define your own rule-set or use one of the recipes` something to that effect?

docs/README.md
```diff
@@ -19,417 +17,498 @@
...
-There may be certain formats and structure you want your Github issues to adhere to. Mergeable allows you to configure this and will create a comment with a list of suggested improvements to your Issue.
+```yml
+version: 2
+mergeable:
+  - when: {{event}}, {{event}} # can be one or more
+    validate:
+      # list of validators. Specify one or more.
+      - do: {{validator}}
+        {{option}}: # name of an option supported by the validator.
+          regex: {{expression}} # where regex is a nested option
```
shine2lay: just a thought: wouldn't `{{sub-option}}` be better than `regex` ?

shine2lay: i don't agree with the comment that says `regex is a nested option` because `regex` is not one of the valid `option` so technically you are not having it `nested` here

docs/README.md
```diff
@@ -19,417 +17,498 @@
...
+    pass: # list of actions. Specify one or more. Omit this tag if no actions are needed.
```
shine2lay: i think it might be good to explicitly say `list of actions to be taken if all the validators returned true`
and there is also a hierarchy here which may not be obviously stated. Which are (in highest to lowest order)
If some validators returned error : do `error` action 
If some validators failed and no errors: do `fail` action
If all validator passed : do `pass` action

docs/README.md
```diff
@@ -19,417 +17,498 @@
...
+      - do: {{action}}            
+    error: # list of actions. Specify one or more. Omit this tag if no actions are needed.
+      - do: {{action}}      
+```
 
-- Notify the author and collaborators when an issues does not adhere to certain formats in the title.
+Take a look at some [example recipes](#examples).
 
-- Notify when projects and milestones are not associated.
+> ☝ **NOTE:** Earlier versions used a [different set of convention](version1.md#configuration). Don't worry, It is fully compatible.
```
shine2lay: Similar to my comment in the v1 config doc, i think we should state that it may be not be fully support in the future versions so we are not always obligated to keep that config forward, which maybe difficult if the system gets more and more complex.

docs/README.md
```diff
@@ -19,417 +17,498 @@
...
 
-- Kanban WIP limits. Limit the number of open Pull Requests by author.
+```yml
+- do: approvals
+  min:
+    count: 2 # Number of minimum reviewers. In this case 2.
+    message: 'Custom message...'
+  required:
+    reviewers: [ user1, user2 ]   # list of github usernames required to review
+    owners: true # accepts boolean. When true,  the file .github/CODEOWNER is read and owners made required reviewers
```
shine2lay: I think we should explicitly say that this won't work with the public mergeable version right now because we don't have permission to read `.github/CODEOWNER` file

docs/README.md
```diff
@@ -19,417 +17,498 @@
...
 
-- Better GitHub Projects automation. i.e. When an issue is assigned, the card automatically moves from column 1 (i.e. `Backlog` if configured as such) to column 2 (i.e. `In Progress` if configured as such) in the kanban board.
+### assignee
 
-- Improved Project Management Integrations (two way): Clubhouse, Pivotal Tracker, Jira, Trello
+```yml
+- do: assignee
+  max:
+    count: 2 # maximum number of assignees
+    message: 'test string'
```
shine2lay: i would advocate to use `There should not be more than 2 assignees` to make it clearer about what the message is for

docs/README.md
```diff
@@ -19,417 +17,498 @@
...
+  min:  
+    count: 2 # min number of assignees
+    message: 'test string'
```
shine2lay: same here as above , `There should be at least 2 assignees`

docs/README.md
shine2lay: would be nice if add a comment saying that the `message` field is optional

docs/version1.md
```diff
@@ -0,0 +1,276 @@
+Configuration of the rulesets can be done in two ways: Simple or advanced.
```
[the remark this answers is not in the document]
jusx: As time goes on we can sunset v1 -- I don't intend to rework that to support new validators and actions. The old stuff will continue for the forseable future but once it gets in the way of progress we'd not support it.

docs/README.md
jusx: I think that we don't have enough collection of existing examples to choose from. Hence I feel we can leave it this way. I did contemplate on using the word rule-set but that is the same as recipe...

docs/README.md
jusx: It works for public repos and I think we should set the permissions shortly after we turn flex on. I think the time span of that is small enough to not mention it. Otherwise we'd have to come back to update it. For now I view it as a bug.

docs/README.md
jusx: Good suggestion. I updated with your proposal.

docs/README.md
jusx: Added annotation to say that it is optional.

docs/README.md
jusx: I updated the verbage. LMK what you think.

docs/README.md
jusx: Updated as you suggested. LMK what you think.

docs/README.md
jusx: Updated as per your suggestion. Make sense.

docs/deploy.md
```diff
@@ -0,0 +1,34 @@
...
+- Webhook URL - **Your webhook url for listening to events** (local deployments you can use [smee.io](smee.io))
+- Webhook secret - **Your generated webhook seceret**
+
+**Permissions:**
+- Checks - **Read & Write**
+- Issues - **Read & Write**
+- Repository metadata - **Read Only**
+- Pull requests - **Read Only**
+- Commit Statuses - **Read & Write**
+- Single File - **Read-only**
```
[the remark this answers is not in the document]
jusx: Added repository content read only permission.

docs/README.md
```diff
@@ -57,18 +57,18 @@ mergeable:
       # list of validators. Specify one or more.
       - do: {{validator}}
         {{option}}: # name of an option supported by the validator.
-          regex: {{expression}} # where regex is a nested option
-    pass: # list of actions. Specify one or more. Omit this tag if no actions are needed.
+          {{sub-option}}: {{value}} # an option can have have one or more sub-options.
```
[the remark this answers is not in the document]
jusx: OK, I changed it to `will` for softer language.

PR #139, opened 2019-01-14
fix approvals and no_empty enabled option

shine2lay:
Goal
---
- Fix approval option always running `required` sub-option
- if no_empty.enabled is false, display a proper message instead of throwing errors

### DIFF
--- a/lib/validators/options_processor/options/no_empty.js
+++ b/lib/validators/options_processor/options/no_empty.js
@@ -7,8 +7,15 @@ class NoEmpty {
 
     const enabled = filter['enabled']
     let description = filter['message']
-    if (!enabled) {
+    if (!enabled && enabled !== false) {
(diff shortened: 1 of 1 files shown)


lib/validators/options_processor/options/no_empty.js
```diff
@@ -7,10 +7,17 @@ class NoEmpty {
...
     let description = filter['message']
-    if (!enabled) {
+    if (!enabled && enabled !== false) {
       throw new Error(ENABLED_NOT_FOUND_ERROR)
     }
 
+    if (enabled === false) {
+      return {
+        status: 'pass',
+        description: 'No_empty option is not enabled, as such this check did not run'
```
jusx: I propose that we codify `no_empty` -- pretty sure that works in checksuite and use the word "validate" instead of "check" since we call them validators and if I remember correctly "validate" is the term we are using in the README.

PR #144, opened 2019-01-16
display the proper project names

shine2lay:
Context
---
Currently the input for project validators is `[Object]`

### DIFF
--- a/lib/validators/project.js
+++ b/lib/validators/project.js
@@ -52,7 +52,6 @@ class Project extends Validator {
 
       isMergeable = issuesPrCloses.some(issue => idsFromCards.includes(issue))
     }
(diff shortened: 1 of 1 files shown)


lib/validators/project.js
```diff
@@ -52,7 +52,8 @@ class Project extends Validator {
 
       isMergeable = issuesPrCloses.some(issue => idsFromCards.includes(issue))
     }
-    return consolidateResult([constructOutput(validatorContext, projects, validationSettings, {
+    console.log('projects', projects)
```
jusx: I think you intended for this to be removed. Yes?

shine2lay: i'll remove it when i get home later today and merge the PR

PR #160, opened 2019-01-26
docs: add and/or options to read me

### LINKED ISSUE #76: Document Complex Logic for Labels and Other Checks
Update README with examples of Complex logic checks as per the PR #75

shine2lay:
Goal
---
Document examples of and/or options

closes #76

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -301,7 +301,35 @@ Supported events:
 ### Advanced Logic
 Validators can be grouped together with `AND` and `OR` operators:
 
-TODO: Show examples.
+```yml
+- do: description
+  and:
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -301,7 +301,35 @@ Supported events:
...
+    - must_include:
+        regex: 'Goal'
+        message: 'Please include the goal of the PR'                 
+```
+
+`AND` and `OR` operators can also be nested
+
+```yml
+- do: label
+ or:
```
jusx: This looks like the wrong indentation. However, it may not be wrong yml -- just does not line up when compared to the example above.

PR #169, opened 2019-02-07
fix(flex): before validate invocation

### LINKED ISSUE #168: Checks is created but never completed due to beforeValidate invocation that shouldn't be
Events are triggering check creation when it shouldn't. It shouldn't create checks because it was an event that is not specified in the configuration. This is very problematic because the event triggers check creation and never sends an update for that check with a `complete` status. 

Hence, from a user perspective it seems like Mergeable hung.

## Steps to Reproduce

1. Create a basic configuration like this:
```yml
version: 2
mergeable:
  - when: pull_request.*
    validate:
      - do: title
        must_exclude:
          regex: wip|work in progress|do not merge
```
2. Create a pull request.
3. Submit a pull request review by commenting, or approving the PR. 

### Current Behavior
Mergeable checks starts processing but never completes.

### Expected Behavior
Mergeable checks should never run because I didn't specify any events in the recipe for the event `pull_request_review`.

## Workaround
Add all relevant PR events in the recipe:
```yml
```

jusx:
Check if the event triggered is relevant before invoking `beforeValidate` on any `Action`s

fixes #168 

## Changes
- Added unit tests for checking proper before and after validate invocations.
- Added checks for all recipes and it's events before invoking beforeValidate.

### DIFF
--- a/lib/flex.js
+++ b/lib/flex.js
@@ -53,6 +53,8 @@ const formatErrorSummary = (errors) => {
 const processWorkflow = async (context, registry, config) => {
   // go through the settings and register all the validators
   config.registerValidatorsAndActions(registry)
+
+  // do pre validation actions
   await processPreActions(context, registry, config)
 
   // process each rule found in configuration
--- a/lib/flex.js
+++ b/lib/flex.js
@@ -52,9 +52,7 @@ const formatErrorSummary = (errors) => {
 
 const processWorkflow = async (context, registry, config) => {
   // go through the settings and register all the validators
-  await config.registerValidatorsAndActions(registry)
-
-  // do pre validation actions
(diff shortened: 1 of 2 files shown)


lib/flex.js
```diff
@@ -52,7 +52,7 @@ const formatErrorSummary = (errors) => {
 
 const processWorkflow = async (context, registry, config) => {
   // go through the settings and register all the validators
-  await config.registerValidatorsAndActions(registry)
+  config.registerValidatorsAndActions(registry)
```
jusx: I removed the `await` keyword here because the function `registerValidatorsAndActions` is not even async. It is redundant and confusing IMO.

PR #170, opened 2019-02-10
feat: request_review + test

### LINKED ISSUE #129: Advanced reviewers settings
It would be nice to have mergeable (automagically) request a review from all people in the reviewers list.

shine2lay:
Goal
---
Add the action to request_review.
closes #129 

usage : 
```
- do : request_review
  reviewers: ['name1', 'name2']
```
An API call will only be made if the reviewers are not requested already.

Note: The reviewers must be collaborator, otherwise, github api will throw error


lib/actions/request_review.js
```diff
@@ -0,0 +1,42 @@
+const { Action } = require('./action')
+const _ = require('lodash')
+
+const createRequestReview = async (context, number, reviewers) => {
+  return context.github.pullRequests.createReviewRequest(
+    context.repo({ number, reviewers })
+  )
+}
+
+class Comment extends Action {
```
jusx: Class name should be `RequestReview` here. :-)

lib/actions/request_review.js
```diff
@@ -0,0 +1,42 @@
...
+  constructor () {
+    super()
+    this.supportedEvents = [
+      'pull_request.opened'
```
jusx: I'm curious if there is a reason for only supporting `opened` ? Seems like it can be `pull_request.*` and `pull_request_review.*`.  Thoughts?

lib/actions/request_review.js
```diff
@@ -0,0 +1,42 @@
+const { Action } = require('./action')
+const _ = require('lodash')
+
+const createRequestReview = async (context, number, reviewers) => {
+  return context.github.pullRequests.createReviewRequest(
```
jusx: Should there be a try catch here so that we can catch the error for when the reviewer is not a collaborator? I feel we should handle it if the request will not complete properly upstream in flex when it this call throws an error. When that happens, we can end up with additional error logs which becomes noise since it is a "known" error.

lib/actions/request_review.js
shine2lay: we don't want to fire request review every time we have an pull_request event.

shine2lay: can also cause https://developer.github.com/v3/#abuse-rate-limits

lib/actions/request_review.js
shine2lay: Done

lib/actions/request_review.js
shine2lay: I've fixed the code to check to make sure already requested reviewers are not being requested again so i've added this to all pull_request events

lib/actions/request_review.js
shine2lay: I've instead chosen to check if the users are collaborator first before requesting review

shine2lay:
@jusx Ready to be reviewed again

weyert:
Exciting how can this PR be progressed?

shine2lay:
@weyert would love some help with manual testing! since @jusx is busy, i guess when the manual testing passes, i can merge it in without @jusx approval.

weyert:
Sure, I can give a shot at it. 

My understanding is that when you create  pull request it should always mark the listed users as reviewers for the pull request. Any other test cases?

shine2lay:
@weyert Yeah, and It would be nice to also test the behavior when the requested_reviewer is not a collaborator. The correct behavior in that case should be that they are not requested for a review and simply skipped.

weyert:
Sweet, I will give it a shot tomorrow :)

weyert:
I have mreged this PR in my custom forked repo and I am getting the error:
```
{ Error: Cannot find module '../validators/request_review'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at rule.validate.forEach.validation (/app/lib/configuration/configuration.js:50:29)
    at Array.forEach (<anonymous>)
    at settings.forEach.rule (/app/lib/configuration/configuration.js:46:23)
    at Array.forEach (<anonymous>)
    at Configuration.registerValidatorsAndActions (/app/lib/configuration/configuration.js:44:19)
    at processWorkflow (/app/lib/flex.js:56:10)
    at executeMergeable (/app/lib/flex.js:38:11) code: 'MODULE_NOT_FOUND' }
15:46:37.849Z ERROR probot: Validators have thrown Error: Cannot find module '../validators/request_review'
  Error: Validators have thrown Error: Cannot find module '../validators/request_review'
      at settings.forEach.rule (/app/lib/configuration/configuration.js:56:15)
      at Array.forEach (<anonymous>)
      at Configuration.registerValidatorsAndActions (/app/lib/configuration/configuration.js:44:19)
      at processWorkflow (/app/lib/flex.js:56:10)
      at executeMergeable (/app/lib/flex.js:38:11)
```

shine2lay:
@weyert that's weird, do you have `request_review.js` in your `validators` folder? if you have code from this PR, that error shouldn't show up.

weyert:
I have merged this PR but no file `request_review.js` in `validators` only in the `actions` folder

lib/actions/request_review.js
```diff
@@ -0,0 +1,64 @@
...
+      'pull_request_review.submitted',
+      'pull_request_review.edited',
+      'pull_request_review.dismissed',
+      'pull_request.labeled',
+      'pull_request.milestoned',
+      'pull_request.demilestoned',
+      'pull_request.assigned',
+      'pull_request.unassigned',
+      'pull_request.unlabeled',
+      'pull_request.synchronize'
```
jusx: Would it make sense to specify `pull_request.*` instead?

docs/README.md
```diff
@@ -460,6 +460,24 @@ Supported events:
...
 ```
+
+### request_review
+Creates comments in issues and/or pull requests depending on the event specified in the `when` tag.
+
+```yml
+- do: request_review
+  reviewers: ['name1', 'name2']
+```
+An API call will only be made if the reviewers are not requested already.
```
jusx: I would change it to  "This is only enforced if the reviewers who are not already requested."

shine2lay: addressed had to change the wording a bit

lib/actions/request_review.js
shine2lay: done

PR #175, opened 2019-02-13
fix(approvals): Should not fail when reviewer approved and later comment

### LINKED ISSUE #174: Checks when specifying OR approvals in the config doesn't seem to consider all approvers.
## To Reproduce
1. Create a recipe similar like this:
**copied from https://github.com/jusx/mergeable/issues/171**
```yaml
version: 2
mergeable:
  - when: pull_request.open, pull_request_review.*
    validate:
      - do: approvals
        or:
          - required:
              reviewers: ['user_one']
          - required:
              reviewers: ['user_two']
```

1. Open a PR.
2. Have 3 approvers, 2 within the or statements ("user_one" and "user_two") and 1 ("user_three") that isn't.

### Actual
The Mergeable check in the PR only considers the 1 that isn't on the list.

### Expected
The Mergeable check should consider all 3 and approve.

## Notes
This issue is occurring in a private repo

jusx:
fixes #174 

## Changes
- Added more tests to `OR` operands to improve coverage.
- Fix the logic to find approvers in the PR reviews. It should not take into account the comment after an approval is made. Since it does not approve or reject a PR we filter it out all together.
- Add test to cover all edge cases for review states.

### DIFF
--- a/__tests__/validators/approvals.test.js
+++ b/__tests__/validators/approvals.test.js
@@ -105,7 +105,7 @@ test('mergeable is false if required member(s) has not approved', async () => {
   expect(validation.validations[0].description).toBe('approval: userC required')
 })
 
-test('mergeable is false if required user has not approved', async () => {
+test('mergeable passes if required user has not approved', async () => {
   const approval = new Approval()
   const reviewList = [
     {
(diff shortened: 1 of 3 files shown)


__tests__/validators/options_processor/options/or.test.js
```diff
@@ -1,7 +1,7 @@
 const or = require('../../../../lib/validators/options_processor/options/or')
 
 const validatorContext = {
-  name: 'label',
+  name: 'OptionName',
```
jusx: Changed it to a generic name to clarify that value of the string does not matter in this fixture.

__tests__/validators/approvals.test.js
```diff
@@ -105,7 +105,7 @@ test('mergeable is false if required member(s) has not approved', async () => {
   expect(validation.validations[0].description).toBe('approval: userC required')
 })
 
-test('mergeable is false if required user has not approved', async () => {
+test('mergeable passes if required user has not approved', async () => {
```
jusx: Updated the description of the test here as it was incorrect.

PR #178, opened 2019-02-13
fix(approvals): finding approved reviewers should properly ignore comments

jusx:
Should take care of #174 

## Changes
- The GH API payload returns the states in past tense, we coded as a verb (action) as documented for submission. Changed to evaluating past tense of the verb for `COMMENT`


shine2lay:
good catch

PR #185, opened 2019-03-14
fix: Change license in package.json to match LICENSE file

IllegalCactus:
Sets the license in package.json so that it matches the one contained in
the LICENSE file.

The license identifier is based on:
https://spdx.org/licenses/


jusx:
Thanks!

PR #187, opened 2019-03-16
fix: Remove unecessary condition from no_empty.js

IllegalCactus:
At some point, logic was added in this file that checks whether `enabled
=== false`, and if so returns early with a warning.

As such, the validation logic that happens below this doesn't need to check
this condition again.


IllegalCactus:
Sorry for the wave of PRs - I figured it was best to keep these small things separate. You may or may not want to merge these but they're just small things I noticed while reading through the code.

jusx:
@IllegalCactus Hey, thanks for submitting these. The PRs are concise and well separated. Please feel free to submit anything you think make sense!

PR #188, opened 2019-03-23
docs: Update deploy.md docs

IllegalCactus:
Update the deployment docs a bit, to clarify a few things about running locally. I had a few issues running this app locally at first, such as with the webhook secret (the GitHub app page mentions it's optional, but the probot wouldn't work locally without one set).

Though this could just be my inexperience with Probots in general, and I am wary of just replicating the existing documentation that GitHub has :slightly_smiling_face:

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -2,12 +2,14 @@
 
 If you would like to run your own instance of this plugin, you can do so by forking this repo and deploying it to your own servers or run it locally.
 
+The [Probot deployment guide](https://probot.github.io/docs/deployment/) describes this as well.
+
 [Create a GitHub App](https://github.com/settings/apps/new) and configure the permissions & events with the following:
 
 **Settings:**
 - GitHub app name - **Your app name**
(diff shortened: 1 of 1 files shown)


docs/deploy.md
```diff
@@ -17,19 +19,35 @@ If you would like to run your own instance of this plugin, you can do so by fork
...
+
 ## Running Locally
 1. Clone the forked repository on to your machine
 2. Globally install smee-client from with npm ```npm install -g smee-client```
-3. Export all the variables required based on the ProBot deployment guide above
-4. Run smee in your terminal by using the ```smee``` command
-5. Run npm start in your local repository
+3. Go to [smee.io](https://smee.io) and create a new webhook OR use the cli by
+   running the `smee` command.
+3. Export all the variables required based on the ProBot deployment guide above:
```
jusx: @IllegalCactus While GH markdown convert these numbered bullet points accurately, it will be nice to keep the numbers in the correct running order. In this case it should be `4` here.  Alternatively, I'm open to have this all changed to just `1` :-)

IllegalCactus: Woops, good point. Have fixed this.

PR #192, opened 2019-03-30
feat: Add PR size validator

IllegalCactus:
Hi, I noticed that in the README there was mention of adding a way to validate the size of PRs. This sounded like a useful feature, and is something I was interested in using the bot for at my work.

This PR is my attempt at adding this. I've tested it out locally, and it seems to work OK.

I'm still not very familiar with the codebase, so this might not be the right way to approach this - let me know what you think.

It's also quite simple at the moment: it only lets you specify a maximum size for the total, and only counts added or modified files. Let me know if you'd prefer it to be more flexible or if this is OK as a start (it should be fairly easy to change the behavior either way).

Apologies for the (ironically) somewhat large PR. I've tried to keep it small, but there are a couple of tiny fixes/refactors I can take out if you'd prefer to make this smaller.

Screenshot of local test run:
failure_screenshot

Full description of changes from the commit:

> Adds a new `size` validator that allows for validating the size of PRs
> in terms of changes. Currently, this only lets you validate that the
> total no. of changed lines is below a maximum using the `max` option.
> 
> It also supports an `ignore` setting to allow excluding certain files
> from the total size (e.g. for ignoring automatically generated files
> that increase the size a lot).
> 
> Things it does not support but could be added in future:
>   - Using different options than `max`
>   - Validating additions and deletions separately (e.g allowing 300
>     additions max, 600 deletions max)
>   - Including entirely deleted files in the total (can very easily be
>     added)
> 
> Other changes that were needed to get this working:
>   - Update `max` option to support integers as well as Arrays

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -14,7 +14,7 @@
   </a>
   <a href="https://circleci.com/gh/jusx/mergeable">
     <img src="https://circleci.com/gh/jusx/mergeable.svg?style=shield">
-  </a>  
+  </a>
   <a href="https://codecov.io/gh/jusx/mergeable">
     <img src="https://codecov.io/gh/jusx/mergeable/branch/master/graph/badge.svg">
   </a>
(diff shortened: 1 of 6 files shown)


docs/README.md
```diff
@@ -462,7 +462,30 @@ Validate pull requests for mergeability based on content and structure of your P
...
+  <p>
+
+  ```yml
+  version: 2
+  mergeable:
+    - when: pull_request.*
+      validate:
+        - do: size
+          ignore: ['ignore_me.js']
+          max:
```
shine2lay: Can i suggest having 
```
lines:
  max: ...
``` 
instead? i can imagine a case where size of an PR is determined by the number of files affected as well.
The feature to check PR size by files could be implemented later.

IllegalCactus: Ah, good idea! I've updated it to follow that structure instead.

shine2lay:
@IllegalCactus Hey! sorry it took me so long to merge it, i was traveling overseas. I just merged it, it should be automatically deployed to production in a about 10 mins.

IllegalCactus:
No worries, thanks for the review @shine2lay 🙂

jusx:
@IllegalCactus Appreciate the PR and the code coverage. :-)

PR #193, opened 2019-04-06
docs: Add 'size' validator to list of validators in README

IllegalCactus:
I realised that in https://github.com/jusx/mergeable/pull/192 , I only added an example snippet of this validator, but didn't add a description of it to the list of validators at the top.

This should document it a little clearer I hope.


jusx:
Looks good @IllegalCactus. Thanks for the clarifying documentation.

PR #196, opened 2019-04-15
feat: Check Re-run

### LINKED ISSUE #152: Re-run check is not working
The [re-run check](https://user-images.githubusercontent.com/1620565/51492409-7c55ba80-1d90-11e9-87fc-ac426a7db645.PNG) capability is not working.

Once you click on it, it'll say that it's pending and it'll stay like that for ever until an event occurs (e.g. a label is added, etc). Only then is the mergeable check re-run.

jusx:
## Goals
- closes #152 

Adds the ability to re-run `checks`:
image

Note: This PR does not address the `Re-run all checks` link.

## Changes
- Added interceptor middleware 
- Added check-rerun interceptor to update context during a check rerun so that validation can be run with the right payload.

### DIFF
--- a/__fixtures__/helper.js
+++ b/__fixtures__/helper.js
@@ -89,7 +89,8 @@ module.exports = {
           },
           getReviews: async () => {
             return { data: (options.reviews) ? options.reviews : [] }
-          }
+          },
+          get: jest.fn()
         },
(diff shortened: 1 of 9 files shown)


jusx:
Yes. It's the same strategy applied with pulls and milestoned.

PR #200, opened 2019-04-17
feat: Update probot package

IllegalCactus:
Update the 'probot' npm package to the latest version.

This requires a number of small changes to deprecated methods in the github
context API. A couple of these were breaking, but most were just
deprecation warnings.

The main motivation for this was to fix a bug where the PRIVATE_KEY
environment variable was not being correctly recognised if it was base64
encoded, thus preventing deployment of the bot. base64-encoded keys are
supported in later versions of the probot lib:

https://probot.github.io/docs/configuration/

Support for this was added in 7.1.0:

https://github.com/probot/probot/pull/624

### DIFF
--- a/__fixtures__/helper.js
+++ b/__fixtures__/helper.js
@@ -48,7 +48,7 @@ module.exports = {
       github: {
         repos: {
           createStatus: () => {},
-          getContent: ({ path }) => {
+          getContents: ({ path }) => {
             return new Promise((resolve, reject) => {
               if (path === '.github/mergeable.yml') {
                 throwNotFound()
@@ -80,7 +80,7 @@ module.exports = {
           }
         },
         pullRequests: {
-          getFiles: () => {
+          listFiles: () => {
             if (_.isString(options.files && options.files[0])) {
               return {
                 data: options.files.map(
@@ -97,8 +97,12 @@ module.exports = {
               return { data: options.files && options.files }
             }
           },
-          getReviews: async () => {
-            return { data: (options.reviews) ? options.reviews : [] }
+          listReviews: {
+            endpoint: {
+              merge: async () => {
+                return { data: (options.reviews) ? options.reviews : [] }
+              }
+            }
           },
           get: jest.fn()
         },
@@ -106,18 +110,18 @@ module.exports = {
           return fn.then(cb)
         }),
         projects: {
-          getRepoProjects: () => {
+          listForRepo: () => {
             return { data: (options.repoProjects) ? options.repoProjects : [] }
           },
-          getProjectColumns: () => {
+          listColumns: () => {
             return { data: (options.projectColumns) ? options.projectColumns : [] }
           },
-          getProjectCards: () => {
+          listCards: () => {
             return { data: (options.projectCards) ? options.projectCards : [] }
           }
         },
         issues: {
-          getIssueLabels: () => {
+          listLabelsOnIssue: () => {
             return { data: (options.labels) ? options.labels : [] }
           },
           get: () => {
@@ -142,12 +146,12 @@ module.exports = {
   },
 
   mockConfigWithContext: (context, configString, options) => {
-    context.github.repos.getContent = () => {
+    context.github.repos.getContents = () => {
       return Promise.resolve({ data: {
         content: Buffer.from(configString).toString('base64') }
       })
     }
-    context.github.pullRequests.getFiles = () => {
+    context.github.pullRequests.listFiles = () => {
       return Promise.resolve({
         data: options && options.files ? options.files.map(file => ({ filename: file, status: 'modified' })) : []
       })
--- a/lib/actions/comment.js
+++ b/lib/actions/comment.js
@@ -1,9 +1,9 @@
 const { Action } = require('./action')
 const populateTemplate = require('./handlebars/populateTemplate')
 
-const createComment = async (context, number, body) => {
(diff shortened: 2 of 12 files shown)


__fixtures__/helper.js
```diff
@@ -97,8 +97,12 @@ module.exports = {
...
             }
           },
-          getReviews: async () => {
-            return { data: (options.reviews) ? options.reviews : [] }
+          listReviews: {
+            endpoint: {
+              merge: async () => {
+                return { data: (options.reviews) ? options.reviews : [] }
+              }
+            }
```
IllegalCactus: The arguments to `context.github.paginate` changed - the deprecation warning I got was:

```
Error: .paginate(promise) is deprecated. Use .paginate(endpointOptions) instead.

For example, instead of

    context.github.paginate(context.github.issues.getAll(context.repo())

do

    context.github.paginate(context.github.issues.getAll.endpoint.merge(context.repo())

Note that when using the new syntax, the responses will be mapped to its data only by default.
```

lib/actions/comment.js
```diff
@@ -1,9 +1,9 @@
 const { Action } = require('./action')
 const populateTemplate = require('./handlebars/populateTemplate')
 
-const createComment = async (context, number, body) => {
+const createComment = async (context, issueNumber, body) => {
   return context.github.issues.createComment(
-    context.repo({ number, body })
+    context.repo({ issue_number: issueNumber, body })
```
IllegalCactus: `number` was deprecated in favor of `issue_number`.

Same goes for the `number` -> `pull_number` changes in other files.

IllegalCactus:
Hi, so for context I've been trying to deploy this bot in my organisation but we've hit an issue where our base64 encoded `PRIVATE_KEY` won't work. We narrowed the issue down to the version of `probot` used in mergeable.

I appreciate it's a big version bump so let me know what you think - I'm happy to do this differently (e.g. upgrading to something smaller).

I've tested this out locally by using a private test repo, and making a `mergeable.yml` with all the validator examples in the README. I'm hoping that this tested most of the funcionality, but I'm a bit worried I've missed something as the unit tests mock the API so it's hard to tell for sure. If someone else wants to test this as well, that'd probably be best :)

jusx:
Looks like almost all of the changes are due to the API changes from the underlying `ocktokit/rest.js` -- At a glance, it looks good. I'll take a deeper look at this upgrade over the next few days as well as testing it on my test repo. 

@shine2lay would be great if you do some manual testing on this as well.

Hopefully in the future we have a way to automate all of these e2e tests. First thing that comes to mind:

- have test repo(s) with PRs each with their own yml for different scenarios. 
- Issues can be tested the same way.
- functional tests triggers PR check re-runs, validate the results 
- functional tests triggers event on issues by making API calls, validate the results.

A ton of work. Thoughts on a better way to automate this?

IllegalCactus:
I was thinking some kind of e2e/integration tests would be great - I don't have any good suggestions atm other than what you described above though. It does seem like it would be a lot of work to setup.

I did a quick Google to see if others had solved this somehow but couldn't find anything.

jusx:
@IllegalCactus I have tested all the actions and validators with the upgrade. I found some problems:

1. We intercept and update the context event property depending on the scenario for every webhook event. In the new version the event property is only a getter. I have opened a PR to ensure upward compatibility #204 -- once we merge that, then this PR is in the clear for the interceptors.

2. probot-scheduler needs to be updated. The version we have in package.json will not work with probot 9.x (https://www.npmjs.com/package/probot-scheduler) -- I'm ok with upgrading to the beta. I've looked through the changes. It is breaking and it's minimal.

3. The `project` validator needs to be updated with `context.github.projects.listForRepo`, `projects.listCards`, `projects.listColumns` respectively.

I looked through the changes in probot. This is actually a one major version bump. Probot 8.x and 9.x is actually the same. And having gone through the release notes, I think that this PR is not that risky.

If you can take care of `2.` and `3.` in this PR we'd be in a good place. I chatted with @shine2lay -- he will do a round of testing this weekend. Just to be on the safe side.

IllegalCactus:
Thanks for the thorough review and test! Have updated to address points 2. and 3.

I've upgraded to `probot-scheduler@^2.0.0-beta.1`  - I assume that's the beta version you meant?

I also tested the `project` validator locally as well (I'd missed that one out previously). Seems to work OK for me now.

I also fixed one more deprecation warning I spotted regarding `error.code` vs `error.status` (in `owners.js`). There are a couple more deprecation warnings regarding `result.data.repositories` but that seem to be occurring inside an octokit pagination plugin? So not sure if I can do anything about that:

```
{ Deprecation: [@octokit/rest] "result.data.repositories" is deprecated. Use "result.data" instead
    at Array.get (mergeable/node_modules/@octokit/rest/plugins/pagination/normalize-paginated-list-response.js:59:39)
    at mergeable/node_modules/probot/lib/apps/stats.js:93:70
    at iterator.next.then.result (mergeable/node_modules/@octokit/rest/plugins/pagination/paginate.js:26:40) name: 'Deprecation' }
```

shine2lay:
@IllegalCactus @jusx 
I found one tiny bug that isn't related to changes in this particular PR, I've opened a PR to fix that in #206 
Other than that, I didn't run into any problem when i was testing.

jusx:
@IllegalCactus Looks good! Thanks for making the changes. 
@shine2lay Thanks for testing this PR as well.

PR #205, opened 2019-04-20
feat: defaults are only loaded when no configuration found

### LINKED ISSUE #194: Default configuration doesn't make sense
Hre's default configuration which is merged if there is no events for `issues` or `pull_requests` keys specified in user-defined config:
```
Configuration.DEFAULTS = {
  label: 'work in progress|do not merge|experimental|proof of concept',
  title: 'wip|dnm|exp|poc',
  stale: {
    message: 'There haven\'t been much activity here. This is stale. Is it still relevant? This is a friendly reminder to please resolve it. :-)'
  }
}
```

Consider PR titled "Implement CSV export". Mergable fails, because it finds "exp".

This default regexp could be improved by checking for keywords only around `^|$|\s`.

jusx:
## Goals
resolves #194 

## Changes
- Only set defaults when no configuration is found in the repos.
- Change the default title regex to be more precise. It makes more sense to only check for `wip` in the beginning of the title.
- Change the default label regex to only look for "work in progress" and "do not merge."

### DIFF
--- a/lib/configuration/lib/consts.js
+++ b/lib/configuration/lib/consts.js
@@ -92,12 +92,12 @@ Settings : {{{displaySettings details.settings}}}
   DEFAULT_PR_VALIDATE: [{
     do: 'title',
     must_exclude: {
-      regex: 'wip|dnm|exp|poc'
+      regex: '^wip'
     }
   }, {
     do: 'label',
     must_exclude: {
-      regex: 'work in progress|do not merge|experimental|proof of concept'
+      regex: 'work in progress|wip|do not merge'
     }
   }, {
     do: 'description',
--- a/__tests__/configuration/configuration.test.js
+++ b/__tests__/configuration/configuration.test.js
@@ -1,4 +1,5 @@
 const Configuration = require('../../lib/configuration/configuration')
+
 const helper = require('../../__fixtures__/helper')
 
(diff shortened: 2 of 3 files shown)


lib/configuration/lib/consts.js
```diff
@@ -92,12 +92,12 @@ Settings : {{{displaySettings details.settings}}}
...
     do: 'title',
     must_exclude: {
-      regex: 'wip|dnm|exp|poc'
+      regex: '^wip'
     }
   }, {
     do: 'label',
     must_exclude: {
-      regex: 'work in progress|do not merge|experimental|proof of concept'
+      regex: 'work in progress|wip|do not merge'
```
rgtk: I would throw `wip` here. Intentions fails with Issue title like `Creating an user wipes all others`. Would keep only verbose one, for demo purposes.

rgtk: \cc @jusx

jusx: @rgtk - This is validating on the `label` and not the `title`. A label named `wip` seem reasonable here. Thoughts?

rgtk: Right. Thanks.

jusx:
I was not done with this PR. Was going to check some things after resolving conflicts.

PR #206, opened 2019-04-21
fix: condition in which reRun is process

shine2lay:
# Context
There is a bug in checkReRun class where the condition is process reRun is not correct.

### DIFF
--- a/__tests__/interceptors/checkReRun.test.js
+++ b/__tests__/interceptors/checkReRun.test.js
@@ -3,12 +3,12 @@ const CheckReRun = require('../../lib/interceptors/checkReRun')
 const Helper = require('../../__fixtures__/helper')
 require('object-dot').extend()
 
-test('return if pre conditions are not met', async () => {
+test('context is not modified if pre conditions are not met', async () => {
   let checkReRun = new CheckReRun()
   let context = Helper.mockContext()
 
   context.event = 'check_run'
-  context.payload.action = 'opened'
+  context.payload.action = 'created'
   Object.set(context, 'payload.check_run.output.text', mockOutput())
   context.github.pullRequests.get.mockReturnValue({ data: { number: 456 } })
   let newContext = await checkReRun.process(context)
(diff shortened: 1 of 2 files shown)


__tests__/interceptors/checkReRun.test.js
```diff
@@ -3,34 +3,58 @@ const CheckReRun = require('../../lib/interceptors/checkReRun')
...
+
+  context.payload.check_run.pull_requests = [{number: 1}]
+  context.payload.check_run.id = 123
+
+  newContext = await checkReRun.process(context)
   expect(newContext.event).toBe('pull_request')
-  expect(newContext.payload.action).toBe('unlabeled')
 })
 
+// test('#possibleInjection', () => {
```
jusx: May I ask the reasons for this entire block to be commented out?

__tests__/interceptors/checkReRun.test.js
```diff
@@ -3,34 +3,58 @@ const CheckReRun = require('../../lib/interceptors/checkReRun')
...
-})
-
-test('#process', async () => {
+test('return if pre conditions are not met', async () => {
   let checkReRun = new CheckReRun()
   let context = Helper.mockContext()
 
   context.event = 'check_run'
-  context.payload.action = 'rerequested'
+  context.payload.action = 'opened'
```
jusx: It might be clearer if we change the action to something that is valid for the event used in the test, i.e. `created` or `completed`

__tests__/interceptors/checkReRun.test.js
```diff
@@ -3,34 +3,58 @@ const CheckReRun = require('../../lib/interceptors/checkReRun')
...
   Object.set(context, 'payload.check_run.output.text', mockOutput())
-  context.payload.check_run.pull_requests = [{number: 1}]
-  context.payload.check_run.id = 123
   context.github.pullRequests.get.mockReturnValue({ data: { number: 456 } })
   let newContext = await checkReRun.process(context)
 
-  expect(newContext.payload.pull_request.number).toBe(456)
+  expect(newContext.event).toBe('check_run')
+
```
jusx: I would very much prefer for the tests to be kept separate:

1. Where `#process` fully executes (as per the original test) 
2. Where the event and action conditions are not met. The inverse of the above and the bug you found.

It would be more readable. Right now this test is checking for both scenarios while the title says "'return if pre conditions are not met'" -- which is rather confusing.

Do you mind keeping the original (commented out) and changing this test to only test for when the conditions are not met as your title implies?

__tests__/interceptors/checkReRun.test.js
shine2lay: done

__tests__/interceptors/checkReRun.test.js
shine2lay: done

__tests__/interceptors/checkReRun.test.js
shine2lay: forgot to uncomment

PR #210, opened 2019-05-07
fix: and update checks in the readMe

shine2lay:
# Context
I tried to copy and paste the checks setting in our readMe and ran into a few invalid configuration error.
Also added the default fail template since if user were to utilize other actions, the defaults are no longer applied so user will have to specify the defaults explicitly

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -398,13 +398,38 @@ Supported events:
 ```
 ### checks
 
+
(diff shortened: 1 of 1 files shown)


shine2lay:
Going to force merge this as jusx is not availabe for review and this only changes readMe

PR #212, opened 2019-05-11
feat: 'assign' action

shine2lay:
Context
Add a new Action : Assign , add a new assignee to an issue or pull_request 
Usage : 
```yml
- do: assign
  assignees: [ 'shine2lay', 'jusx' ] # only array accepted
```


lib/actions/assign.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+const checkAssignee = async (context, issueNumber, assignee) => {
+  const checkResult = await context.github.issues.checkAssignee(
+    context.repo({ issue_number: issueNumber, assignee })
+  ).catch(err => {
+    if (err.status === 404) return { status: 404 }
+  })
+
+  return checkResult.status === 204 ? assignee : null
```
jusx: Would it be more readable if the function is named `isAssigneeValid` and that it returns true or false?

shine2lay: sure, that make sense

PR #214, opened 2019-05-13
feat: Add option for requiring assignee approval

IllegalCactus:
Adds an option to the `approvals.required` config for specifying that all PR assignees are required to approve. When set to true, it automatically adds all assignees to the list of required approvers, similar to how the `owners` option works.

### DIFF
--- a/__tests__/validators/approvals.test.js
+++ b/__tests__/validators/approvals.test.js
@@ -105,7 +105,7 @@ test('mergeable is false if required member(s) has not approved', async () => {
   expect(validation.validations[0].description).toBe('approval: userC required')
 })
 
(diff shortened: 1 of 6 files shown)


IllegalCactus:
Hi, let me know if this is functionality you'd be willing to support / if it makes sense for mergeable in your view. For context we're using mergeable at my work, and wanted functionality like this to ensure that assigned reviewers are required regardless of who else might approve a PR.

I've done some testing of this in our own organisation. So far it's been working as expected, but some testing from others is likely needed :)

shine2lay:
I just tested it, it works great!!

IllegalCactus:
Thanks @shine2lay  :)

Issue #217, opened 2019-06-21
Make size:ignore work with regex or globs


ohager:
In a current project we use react snapshot testing...this may lead to huge changes and thus we need to "disable"  mergeable from time to time, because of rewritten snapshot files.
I figured out, that the the ignore setting in `size` works only with fixed filenames, and does not support filename patterns, e.g. globbing or regex.

Would be nice to have that feature!

IllegalCactus:
@ohager This should be supported now as of: https://github.com/mergeability/mergeable/pull/218

If you’d like to give it a try 🙂

shine2lay:
@ohager i believe @IllegalCactus 's PR has already implemented this feature!! :tada:  :tada: 
I'll go ahead and close this issue, please re-open if you found an issue

PR #218, opened 2019-06-30
feat: Support glob patterns in size validator

IllegalCactus:
The size validator 'ignore' option only allowed specifying absolute file
paths. This adds support for glob pattern matching to allow specifying
patterns to ignore.

This relies on the minimatch library (pulled in as a new dependency) to
perform the glob handling and conversion to regular expressions:

https://github.com/isaacs/minimatch

This addresses https://github.com/mergeability/mergeable/issues/217

### DIFF
--- a/__tests__/validators/size.test.js
+++ b/__tests__/validators/size.test.js
@@ -95,6 +95,61 @@ describe('PR size validator', () => {
     expect(validation.validations[0].description).toBe('PR size is OK!')
   })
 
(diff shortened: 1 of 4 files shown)


IllegalCactus:
I'm a bit unsure about the new dependency. This project has very few at the moment and I'm not sure whether this justifies pulling in a new one. But let me know what you think :)

PR #220, opened 2019-08-01
docs: Fix link.

shields-fn:


shine2lay:
@shields-fn Thanks for the fix!

PR #230, opened 2019-08-29
Fix: add uncaught exception handler for validators

roblabla:
Currently, if a validator throws an exception, the error will be silently lost. This is obviously not ideal. This PR adds an uncaught error handler that will run the actions with a special kind of result that has an error item in the details, hopefully containing a stacktrace.

https://github.com/sunriseos/mergeable-test/pull/2/checks?check_run_id=206717040

While this is an improvement on the current situation, there's another way for users to get an error with no output: if the user has a broken action, we have no way to catch the action, and even if we did, we have no way to communicate the error to the user. 

Screen Shot 2019-08-29 at 12 31 05 PM


lib/flex.js
```diff
@@ -99,7 +122,10 @@ const processPreActions = async (context, registry, config) => {
 const getActionPromises = (context, registry, rule, result) => {
   const actions = rule[result.validationStatus]
   if (actions) {
-    const afterValidateFuncCall = (actionClass, context, action, result) => actionClass.afterValidate(context, action, result)
+    const afterValidateFuncCall = (actionClass, context, action, result) => {
+      console.log(action, result)
```
shine2lay: i don't think this log is needed

lib/flex.js
```diff
@@ -59,18 +61,39 @@ const processWorkflow = async (context, registry, config) => {
...
         .then((result) => {
           const translatedOutput = extractValidationStats(result)
           let promises = getActionPromises(context, registry, rule, translatedOutput)
           if (promises) {
-            Promise.all(promises)
+            return Promise.all(promises)
           }
         })
+        .catch((err) => {
+          console.error('An error occured while executing an action', err)
```
shine2lay: please use `context.log` instead

shine2lay:
tested manually and seems to be working as expected

lib/flex.js
roblabla: That's some leftover debug 😅

roblabla:
Addressed the comments.

shine2lay:
@jusx this is ready to be reviewed. please note that this is not dealing with error where if user makes a typo for validators or actions, we didn't return any check. It took me some time to realize that.

PR #238, opened 2019-12-18
fix: Upgrade probots and lodash packages

IllegalCactus:
Upgrade to the latest versions of these packages. This should hopefully
also fix a prototype pollution vulnerability in lodash (which probots
also pulls in):

https://snyk.io/vuln/SNYK-JS-LODASH-450202


shine2lay:
did a quick smoke check manually and it seems to be working fine

PR #240, opened 2020-01-13
fix: isEventInContext

shine2lay:
### Context
Current implementation of isEventInContext function allows matching for multiple events such as `issues` and `issuesComment`. This leads to a bug where if the config specify both those events as different `when`, unintended actions will be fired.

Solution:
strictly check for the correct event type instead of using `includes`


shine2lay:
@jusx i'll go ahead and merge it

PR #242, opened 2020-01-18
refactor: prepare for e2e testing

shine2lay:
### Context
There are two major changes: 
- Move tests into a folder called `unit` to indicate for unit testing
- Make the mergeable process fully synchronous


shine2lay:
@jusx please pay attention to `flex.js` and `mergeable.js`

lib/flex.js
```diff
@@ -59,18 +59,17 @@ const processWorkflow = async (context, registry, config) => {
   await processPreActions(context, registry, config)
 
   // process each rule found in configuration
-  config.settings.forEach(rule => {
+  for (const rule of config.settings) {
```
jusx: What's the reasoning for changing this at all?

shine2lay: right now, it's just doing `fire and forget`, I don't know how to test that at all since the test would end before the api has been hit. I would love to hear if you have a way to test it

jusx:
It is not clear to me what the folder structure strategy is here. You've moved everything to `unit/` -- is there going to be an `int/` subfolder for integration / e2e? 

Please clarify

shine2lay:
yes, your assumption is correct. I am planning on having `e2e` folder for E2E tests

lib/flex.js
jusx: I'm reffering to the for loop syntax -- is there a difference in doing`forEach` vs. `for` ?

shine2lay: you can't do proper `async/await` in forEach because of the anynomous functions

shine2lay: Unless you collect the calls as Promise and do `await` on the Promises

PR #254, opened 2020-03-17
feat: Remove default config and add new default behaviour

### LINKED ISSUE #252: Installing the app on a repo should not automatically enforce new checks when a config file is not present
I installed Mergeable to a repo on Friday, with the intention of fleshing out some rules this week.

However, a team member submitted an unrelated PR over the weekend that happened to not include a description. Seemingly out of the blue, Mergeable fails and prevents it from being merged. With a little digging, I found that Mergeable loads an [undocumented series of default checks](https://github.com/mergeability/mergeable/pull/205) when no configuration file is present, including one that fails if no description is present.

IMO, this is not a good out-of-box experience. When installing Mergeable, I did not expect, nor was I notified, that I would also be enrolled in a set of tests I had no input on. These default tests are not documented anywhere on the repo's README, either, so I had no way to know they would run.

The default when no config file is present should be to not run anything. If I wish to make use of Mergeable, I will add a config file, with my own rules. Until that happens, Mergeable should not change my repo's behavior.

shine2lay:
### Context
closes #252 

### Implementation
Instead of creating default config, we create a 'success' check for PRs if no config were found. This is in order to communicate to the user that no configs were found.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "scripts": {
     "dev": "LOG_LEVEL=trace NODE_ENV=development nodemon --exec 'npm start'",
     "start": "probot run ./index.js",
-    "test": "jest && standard",
+    "test": "jest __tests__/unit/flex.test.js && standard",
     "test:unit": "LOG_LEVEL=warn jest __tests__/unit/* && standard",
     "test:e2e": "LOG_LEVEL=warn jest __tests__/e2e/* && standard",
     "test-watch": "jest --watch",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "dev": "LOG_LEVEL=trace NODE_ENV=development nodemon --exec 'npm start'",
     "start": "probot run ./index.js",
-    "test": "jest && standard",
+    "test": "jest __tests__/unit/flex.test.js && standard",
```
jusx: Is this intentional?

shine2lay: no, removed

PR #259, opened 2020-03-19
feat: catch non existant validator n action error

### LINKED ISSUE #224: No error feedback when validate.do has an invalid key
If a `validate.do` key includes an invalid value, Mergeable silently does not do anything. It does not report red, yellow, or green at all.

If I set invalid version, Mergeable nicely reports “Invalid configuration”.

If I use `mussst_exclude` instead of `must_exclude`, Mergeable reports this is not supported.

So this is specific to `validate.do` key.

Minimally similar to #154.

This config triggers Mergeable correctly:

```yml
version: 2
mergeable:
  - when: pull_request.opened, pull_request.edited, pull_request.synchronize
    validate:
      - do: description
        must_exclude:
          regex: 'DO NOT MERGE|WIP'
```

Here I changed `description` to `thing` by mistake and Mergeable is silent:

```yml
version: 2
mergeable:
  - when: pull_request.opened, pull_request.edited, pull_request.synchronize
    validate:
      - do: thing
        must_exclude:
          regex: 'DO NOT MERGE|WIP'
```

shine2lay:
closes #224 
output : 
image

@jusx please pay a closer attention to this PR, it has some refactoring involved for better separation

### DIFF
--- a/lib/flex.js
+++ b/lib/flex.js
@@ -1,7 +1,7 @@
 const Configuration = require('./configuration/configuration')
 const extractValidationStats = require('./stats/extractValidationStats')
 const Checks = require('./actions/checks')
-const registry = require('./registry')
+const register = require('./register')
 const interceptors = require('./interceptors')
 const consolidateResult = require('./validators/options_processor/options/lib/consolidateResults')
 const constructErrorOutput = require('./validators/options_processor/options/lib/constructErrorOutput')
@@ -57,7 +57,7 @@ const formatErrorSummary = (errors) => {
 const processWorkflow = async (context, registry, config) => {
   // go through the settings and register all the validators
   try {
-    registry.registerValidatorsAndActions(config.settings, registry)
+    register.registerValidatorsAndActions(config.settings, registry)
   } catch (err) {
     let evt = `${context.event}.${context.payload.action}`
     let checks = new Checks()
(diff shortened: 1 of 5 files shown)


lib/registry.js
```diff
@@ -0,0 +1,52 @@
...
+      throw new Error('Actions have thrown ' + err)
+    }
+  })
+}
+
+module.exports = {
+  registerValidators,
+  registerActions,
+  registerValidatorsAndActions
+}
```
jusx: In general I strongly suggest keeping the order of the parameters consistent. You have register as the first or second parameter interchangeably.

lib/flex.js
```diff
@@ -1,6 +1,7 @@
 const Configuration = require('./configuration/configuration')
 const extractValidationStats = require('./stats/extractValidationStats')
 const Checks = require('./actions/checks')
+const { registerValidatorsAndActions } = require('./registry')
```
jusx: I personally would much prefer doing a require without the scoping. i.e. 

`const registry = require('.registry')`

it is much more obvious that `registry.registerValidatorsAndActions` is part of another module/object then reading `registerValidatorsAndActions` that can be either a function within this same file or coming from somewhere else.

minor comment and personal preference.

lib/registry.js
shine2lay: make sense, good catch

lib/flex.js
shine2lay: done.

lib/registry.js
shine2lay: done

PR #261, opened 2020-03-19
feat: add error catching to and/or option processor

### LINKED ISSUE #250: Errors from options processor is being swallowed up by and/or sub option processor
### Context 
When user provided invalid sub options, the or/and options processor does nothing to bubble it up.
Here is an error from 'begins_with' option
image

and the output : 
image

shine2lay:
closes #250 

output: 
image

### DIFF
--- a/lib/validators/options_processor/options/lib/andOrProcessor.js
+++ b/lib/validators/options_processor/options/lib/andOrProcessor.js
@@ -24,8 +24,15 @@ const andOrProcessor = (validatorContext, input, rule, key) => {
   let isMergeable
   let DEFAULT_SUCCESS_MESSAGE = `All the requisite validations passed for '${key}' option`
   let descriptions = ''
+  let errorExists = false
+  let errorMessage = 'Error occurred: '
 
   validated.forEach(result => {
+    if (result.status === 'error') {
+      errorExists = true
+      errorMessage += result.description
+    }
+
     const resultSuccess = result.status === 'pass'
     if (isMergeable !== undefined) {
       isMergeable = key === 'and' ? isMergeable && resultSuccess : isMergeable || resultSuccess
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "dev": "LOG_LEVEL=trace NODE_ENV=development nodemon --exec 'npm start'",
     "start": "probot run ./index.js",
-    "test": "jest && standard",
+    "test": "jest __tests__/unit/validators/options_processor/options/or.test.js && standard",
```
jusx: Please fix the above. It looks like it's an unintentional commit. May I recommend doing an `alias` for shortcuts in your shell instead of modifying this file?

lib/validators/options_processor/options/lib/andOrProcessor.js
```diff
@@ -24,8 +24,15 @@ const andOrProcessor = (validatorContext, input, rule, key) => {
   let isMergeable
   let DEFAULT_SUCCESS_MESSAGE = `All the requisite validations passed for '${key}' option`
   let descriptions = ''
+  let errorExists = false
+  let errorMessage = 'Error occurred: '
 
   validated.forEach(result => {
+    if (result.status === 'error') {
+      errorExists = true
+      errorMessage += result.description
```
jusx: Do you have a test case for multiple errors from different options and is it possible to use `MD` bullet points or plain text carriage return formatting?

package.json
shine2lay: Sorry, I'll try using `alias`

lib/validators/options_processor/options/lib/andOrProcessor.js
shine2lay: image

lib/validators/options_processor/options/lib/andOrProcessor.js
```diff
@@ -24,8 +24,15 @@ const andOrProcessor = (validatorContext, input, rule, key) => {
   let isMergeable
   let DEFAULT_SUCCESS_MESSAGE = `All the requisite validations passed for '${key}' option`
   let descriptions = ''
+  let errorExists = false
```
jusx: Should this variable be `isErrorExists` ? This would be inline with our naming convention for boolean type variables as declared on line 27. `isMergeable`

PR #262, opened 2020-03-20
feat: And validator

### LINKED ISSUE #191: Dependent validator should support OR & AND operators
The [dependent](https://github.com/jusx/mergeable#dependent) validator should support [OR and AND operators](https://github.com/jusx/mergeable#advanced-logic).

Consider the following for a repo with both Rails and Node:

```yml
- do: dependent
  or: 
    - changed:
      file: package.json
      files: ['package-lock.json', 'yarn.lock']
    - changed:    
      file: “Gemfile”
      files:  ["Gemfile.lock", "Gemfile_next.lock" ]
```

### LINKED ISSUE #216: Validate description if certain files / paths have changed
In our workflow it would be valuable if we could validate the description against a regular expression if certain files in the repository have been changed.

For example if any file in the `module/migrations` directory changes we'd like to enforce a description outlining the change.

shine2lay:
### Context
'And' at Validator level 

Note: contains refactoring of 'Or' validator to reuse the code

closes #216 
closes #191

### DIFF
--- a/lib/validators/lib/andOrValidatorProcessor.js
+++ b/lib/validators/lib/andOrValidatorProcessor.js
@@ -1,0 +1,85 @@
+const register = require('../../register')
+const { getValidatorPromises } = require('../../flex')
+const consolidateResult = require('../options_processor/options/lib/consolidateResults')
+const constructErrorOutput = require('../options_processor/options/lib/constructErrorOutput')
+
+const ERROR_MESSAGE = `Failed to validate because the 'validate' option is missing or empty. Please check the documentation.`
+
+const andOrValidatorProcessor = async (context, settings, registry, validatorName) => {
+  const validatorContext = { name: validatorName }
+
+  if (!Array.isArray(settings) || settings.length === 0) {
+    return consolidateResult(
+      [
+        constructErrorOutput(
+          validatorName,
+          '',
+          settings,
+          ERROR_MESSAGE
+        )
+      ],
+      validatorContext
+    )
+  }
+
+  const rules = { validate: settings }
+
+  try {
+    register.registerValidators(rules, registry)
+  } catch (err) {
+    // Should I avoid a console error and consolidateOutput instead?
+    return consolidateResult(
+      [
+        constructErrorOutput(
+          validatorName,
+          '',
+          settings,
+          'Unsupported validator ' + err
+        )
+      ],
+      validatorContext
+    )
+  }
+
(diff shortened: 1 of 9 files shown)


lib/validators/lib/andOrValidatorProcessor.js
```diff
@@ -0,0 +1,98 @@
+const register = require('../../register')
+const { getValidatorPromises } = require('../../flex')
+const consolidateResult = require('../options_processor/options/lib/consolidateResults')
+const constructErrorOutput = require('../options_processor/options/lib/constructErrorOutput')
+
+const ERROR_MESSAGE = `Failed to validate because the 'validate' option is missing or empty. Please check the documentation.`
```
jusx: I would suggest using a more explicit variable name, i.e. `OPTION_MISSING_ERROR_MESSAGE`

lib/validators/lib/andOrValidatorProcessor.js
```diff
@@ -0,0 +1,98 @@
...
+      validatorContext
+    )
+  }
+
+  const rules = { validate: settings }
+
+  try {
+    register.registerValidators(rules, registry)
+  } catch (err) {
+    // Should I avoid a console error and consolidateOutput instead?
```
jusx: Depends. If you want to know how often this happens, I'd do `log.info`

lib/validators/lib/andOrValidatorProcessor.js
```diff
@@ -0,0 +1,98 @@
...
+  const promises = getValidatorPromises(context, registry, rules)
+
+  if (promises.length === 0) {
+    return consolidateResult(
+      [
+        constructErrorOutput(
+          validatorName,
+          '',
+          settings,
+          ERROR_MESSAGE
```
jusx: Is this right? If wrong validators are specified they'd receive an error, should they still get this error?

shine2lay: this happens when user doesn't specify `validate` options, i added this check above instead

lib/validators/lib/andOrValidatorProcessor.js
shine2lay: removed the comments

lib/validators/lib/andOrValidatorProcessor.js
shine2lay: done

PR #264, opened 2020-03-25
Add log to show which repo are using mergeable

shine2lay:
### Context 
We currently have no means of measuring usage and adding log would help us filter usage if we want it later on. Only create logs when main mergeable logic is ran.

@jusx wdyt on this solution for the meantime?

### DIFF
--- a/lib/flex.js
+++ b/lib/flex.js
@@ -72,6 +72,17 @@ const formatErrorSummary = (errors) => {
 }
 
 const processWorkflow = async (context, registry, config) => {
+  let log = context.log.child({ name: 'mergeable' })
+
+  // log usage of mergeable
+  log.info(`Running Mergeable for: ${context.payload.repository.full_name}`)
+  log.debug({Repo: context.payload.repository.full_name,
+    url: context.payload.repository.html_url,
+    private: context.payload.repository.private,
+    config: config.settings,
+    event: context.event,
+    action: context.payload.action})
+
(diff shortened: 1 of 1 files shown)


lib/flex.js
```diff
@@ -72,6 +72,17 @@ const formatErrorSummary = (errors) => {
...
+  let log = context.log.child({ name: 'mergeable' })
+
+  // log usage of mergeable
+  log.info(`Running Mergeable for: ${context.payload.repository.full_name}`)
+  log.debug({Repo: context.payload.repository.full_name,
+    url: context.payload.repository.html_url,
+    private: context.payload.repository.private,
+    config: config.settings,
+    event: context.event,
+    action: context.payload.action})
```
jusx: 1. Why here and not in the `flex` function in `Mergeable`? There's already similar logging there.
2. Shouldn't it be `info` if you your purpose is to process the logs for usage? We don't always have log level at `debug`

shine2lay: re 1: `flex` function will log out every event we received, IMO it won't provide true usage stats. That's why i log this out after getting a valid config file and is starting to run the core mergeable functions. I didn't remove the log out in `flex` function because i think it might be useful too. 
re 2: I was worried about log storage cost. but now that i think about it, it's probably un warranted. I'll change it

shine2lay:
closing as #268 makes this PR moot

Issue #215, opened 2019-05-21
Block a PR when a changes have been requested


zoispag:
Hi, 

I do not want to enforce approvals, either by user or by number.
But I want to block a PR from being merged, if a reviewer requested changes.

Thus, this will enforce them to re-review a PR after changes have been applied, before merging the PR.

Is this currently possible?

shine2lay:
Hey @zoispag, 
Just to clarify my understanding,
Did you mean PR merging is blocked until when there are no reviewer requested changes? 
So if there are no requested changes, mergeable should pass?

This feature currently does not exists, however, in our implementation of the approvals counts, we check and make sure that the last review is an approval. 
For example, if an reviewer submits a approval, then submits another review requesting changes, this will not count towards approvals required.
If this information doesn't help with your use case, i'll be happy to implement this. shouldn't take too long.

zoispag:
Hey @shine2lay,

I want to allow non-approved PRs to be merged.
I want to blocked PRs that a reviewer requested changes, until the same reviewer approves.

So in short I wan't to block PRs that look like this:

<img width="240" alt="Pasted_Image_21_05_2019__12_30" src="https://user-images.githubusercontent.com/21138205/58089150-62a07480-7bc4-11e9-8aee-c0c8f7e37da5.png">

shine2lay:
@zoispag that make sense!

ndench:
What about when a reviewer has not requested changes, but a review has still been requested from them? Like this:

![2019-08-02_22-14-49](https://user-images.githubusercontent.com/2062388/62369457-08804000-b573-11e9-82e8-b4f1637249fb.png)

You'll notice that branch protection says only 1 approving review is required, but if this particular PR has requested a review from both reviewers, I think it would be good to block it until they both approve. What do you think?

shine2lay:
@ndench interesting, I think this is a valid use case.

al-the-x:
I wouldn't block in the "multiple reviewers requested" case, TBH. I've been on multiple teams that blanket requests for review from folks when only 2 are required. The behavior is difficult to quell.

PR #265, opened 2020-03-26
feat: approval - requested reviewers

shine2lay:
Add a sub option `requested_reviewers` for approvals
for more context, see #215 

Note: this only implement the latter part of #215 

image

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -112,6 +112,7 @@ For convenience, wildcards can be used: `pull_request.*`, `issues.*`, `pull_requ
     reviewers: [ user1, user2 ] # list of github usernames required to review
     owners: true # Optional boolean. When true, the file .github/CODEOWNER is read and owners made required reviewers
     assignees: true # Optional boolean. When true, PR assignees are made required reviewers.
+    pending_reviewer: true #Optional boolean. When true, all the reviewer's approval is 
     message: 'Custom message...'
 ```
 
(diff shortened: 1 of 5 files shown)


docs/README.md
```diff
@@ -112,6 +112,7 @@ For convenience, wildcards can be used: `pull_request.*`, `issues.*`, `pull_requ
     reviewers: [ user1, user2 ] # list of github usernames required to review
     owners: true # Optional boolean. When true, the file .github/CODEOWNER is read and owners made required reviewers
     assignees: true # Optional boolean. When true, PR assignees are made required reviewers.
+    pending_reviewer: true #Optional boolean. When true, all the reviewer's approval is
```
jusx: Missing a space after the `#` -- Also looks like an intentional unfinished sentence?

lib/validators/approvals.js
```diff
@@ -20,6 +21,7 @@ class Approvals extends Validator {
       'pull_request.unlabeled',
       'pull_request.synchronize'
     ]
+    this.supportedEvents.push('block')
```
jusx: Should this be `pull_request.block` ? Thought I couldn't find such an event in the API docs...

shine2lay: this should not be there, it was my mistake

docs/README.md
shine2lay: fixed

PR #269, opened 2020-03-31
fix: new modules to match the code style + add logging module mock for…

shine2lay:
### Context
Some of the new classes introduce wasn't matching the code style of the repo, this PR is meant to fix that plus a few other minor improvements.

### DIFF
--- a/__fixtures__/setup/jestSetUp.js
+++ b/__fixtures__/setup/jestSetUp.js
@@ -1,0 +1,25 @@
+// mock Logger
+jest.mock('../../lib/logger.js', () => {
+  return {
+    init: jest.fn(),
+    create: () => {
+      return {
+        info: jest.fn(),
+        warn: jest.fn(),
+        error: jest.fn(),
+        debug: jest.fn(),
+        fatal: jest.fn()
+      }
+    },
+    logTypes: {
+      EVENT_RECEIVED: 'event_received',
+      CONFIG_INVALID_YML: 'config_invalid_yml',
+      UNKNOWN_ERROR_VALIDATOR: 'unknown_error_validator',
+      UNKNOWN_ERROR_ACTION: 'unknown_error_action',
+      VALIDATOR_PROCESS: 'validator_process',
+      ACTION_BEFORE_VALIDATE_EXECUTE: 'action_before_validate_execute',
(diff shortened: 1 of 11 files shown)


__fixtures__/setup/jestSetUp.js
```diff
@@ -0,0 +1,25 @@
...
+    create: () => {
+      return {
+        info: jest.fn(),
+        warn: jest.fn(),
+        error: jest.fn(),
+        debug: jest.fn(),
+        fatal: jest.fn()
+      }
+    },
+    logTypes: {
```
jusx, line 14: Do we need to repeat these constants here? Is there a way to reference it from `Logger`?

__fixtures__/unit/helper.js
```diff
@@ -41,9 +41,9 @@ module.exports = {
       log: {
         child: (s) => {
           return {
-            debug: (...s) => console.log(`TEST[debug] > ${JSON.stringify(s)}`),
-            info: (...s) => console.log(`TEST[info] > ${JSON.stringify(s)}`),
-            warn: (...s) => console.warn(`TEST[warn] > ${JSON.stringify(s)}`)
```
jusx, line 46: This was intentionally console.log for debugging purposes while working on the unit tests. Any reason for removing it?

__fixtures__/setup/jestSetUp.js
shine2lay: I tried many different ways, it seems jest doesn't allow referencing out-of-scope variables. If you have heard of a method. would love to hear it.

__fixtures__/unit/helper.js
shine2lay: there are debugging log every time we ran tests, that's why i replaced it. I just realized we are using logger module, this function is moot. Also, i forgot to use the `Logger` in one file.

__fixtures__/setup/jestSetUp.js
shine2lay: One solution i can think of is to separate the `logger` from `logTypes`

shine2lay: I implemented it the solution i mentioned

__fixtures__/unit/helper.js
shine2lay: changed the `checkReRun` file to use `Logger` module

PR #271, opened 2020-04-01
refactor: more verbose logs + add more logging places

shine2lay:
### Context
- add `repo` field to logs to make it easier to track usage.
- add logging in more places

### DIFF
--- a/lib/validators/and.js
+++ b/lib/validators/and.js
@@ -13,7 +13,6 @@ class And extends Validator {
   }
 
   async validate (context, validationSettings, registry) {
-    this.logUsage(context, validationSettings)
     return andOrValidatorProcessor(context, validationSettings.validate, registry, 'And')
   }
 }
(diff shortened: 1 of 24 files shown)


lib/validators/and.js
```diff
@@ -13,6 +13,7 @@ class And extends Validator {
   }
 
   async validate (context, validationSettings, registry) {
+    this.logUsage(context, validationSettings)
```
jusx, line 16: Instead of repetitively adding this one line in every single validator, I'd recommend adding a method in the super class `Validator.processValidation` and have `flex` call that instead. Within `processValidation` you'd have two lines of code:

```js
   this.logUsage(context, validationSettings)
   this.validate(context, validationSettings, registry)
```

This way logUsage is always called for all validators and when implementing additional validators you ensure it is logged.

shine2lay: that's a great idea, i'll implement it

shine2lay: done implementing and tests also fixed

PR #272, opened 2020-04-01
feat: Changeset validator

### LINKED ISSUE #184: Approve based on files changed
Hi! I'm not sure this workflow is supported, but we have a repo with two files, `staging.config` and `production.config` which provides the configuration we run in a given environment. Merging a change to `master` automatically triggers the deploy in appropriate environment. I'd like to automatically approve the PR if only the `staging.config` file  file has changed. If the `production.config`  has changed, I'd like to require at least two approvers. Is this possible?

shine2lay:
### Context 
new validator: `changeset` validate against the files in the PR
config 
```
- do: changeset
    no_empty:
       enabled: false # Cannot be empty when true.
       message: 'Custom message...'
    must_include:
       regex: 'yarn.lock'
       message: 'Custom message...'
    must_exclude:
       regex: 'package.json'
       message: 'Custom message...'
    begins_with:
       match: 'A String' # or array of strings
       message: 'Some message...'
    ends_with:
       match: 'A String' # or array of strings
       message: 'Come message...'
    min:
        count: 2 # min number of files in a PR
        message: 'Custom message...'
    max:
       count: 2 # max number of files in a PR
       message: 'Custom message...'
    # all of the message sub-option is optional
```

closes #184

### DIFF
--- a/lib/validators/changeset.js
+++ b/lib/validators/changeset.js
@@ -1,0 +1,31 @@
+const { Validator } = require('./validator')
+
+class Changeset extends Validator {
+  constructor () {
+    super('changeset')
+    this.supportedEvents = [
+      'pull_request.opened',
+      'pull_request.edited',
+      'pull_request_review.submitted',
+      'pull_request_review.edited',
+      'pull_request_review.dismissed',
+      'pull_request.labeled',
+      'pull_request.milestoned',
+      'pull_request.demilestoned',
+      'pull_request.assigned',
+      'pull_request.unassigned',
+      'pull_request.unlabeled',
+      'pull_request.synchronize'
+    ]
+  }
+
+  async validate (context, validationSettings) {
+    // fetch the file list
+    let result = await context.github.pulls.listFiles(context.repo({pull_number: this.getPayload(context).number}))
+    let changedFiles = result.data.map(file => file.filename)
+
+    return this.processOptions(validationSettings, changedFiles)
+  }
+}
+
+module.exports = Changeset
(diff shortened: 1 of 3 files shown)


lib/validators/changeset.js
```diff
@@ -0,0 +1,31 @@
...
+      'pull_request_review.submitted',
+      'pull_request_review.edited',
+      'pull_request_review.dismissed',
+      'pull_request.labeled',
+      'pull_request.milestoned',
+      'pull_request.demilestoned',
+      'pull_request.assigned',
+      'pull_request.unassigned',
+      'pull_request.unlabeled',
+      'pull_request.synchronize'
```
jusx, line 18: Should this be `pull_request.*` ?

jusx:
Instead of merely closing the issue #184 I recommend commenting with a sample configuration or link to the section in the README.

lib/validators/changeset.js
shine2lay: fixed

shine2lay:
> Are you planning on adding this to the README?

added `changeset` to readme

lib/validators/changeset.js
```diff
@@ -0,0 +1,20 @@
+const { Validator } = require('./validator')
+
+class Changeset extends Validator {
+  constructor () {
+    super('changeset')
+    this.supportedEvents = [
+      'pull_request.*'
```
jusx, line 7: Do we need `pull_request_reviews.*` here since you listed that in the README?

shine2lay: that's actually a great catch, fixed

jusx: Did you push your changes? I don't see it.

shine2lay: sorry, i did attempt a push but it was rejected, i should've double checked that the push went thru. I fixed it

PR #273, opened 2020-04-03
feat: Regex flag to must_include and must_exclude sub options

### LINKED ISSUE #226: Option to disable case insensitivity in regexes
Currently mergeable generates regexes with `i` flag. It would be really nice if there was an option to make case sensitive regexes. For me, this is required in a project where I use AngularJS commit convention and the commit's type should be lower case.

shine2lay:
### Context
add the ability for user to define regex_flag to be used in `must_include` and `must_exclude`
leaving 'i' option as default for backward compatibility. 
sample config 
```
- do: labels
        must_exclude:
          regex: 'ABC.js'
          regex_flags: '' # putting '' will remove the default 'i' options 
```
closes #226

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -231,13 +231,13 @@ completely deleted in the PR.
        message: 'Custom message...' # this is optional, a default message is used when not specified.
     must_include:
        regex: '### Goals|### Changes'
-       regex_flag: '' # Optional. Specify the flag for Regex. default is 'i', to disable default use ''
+       regex_flag: 'none' # Optional. Specify the flag for Regex. default is 'i', to disable default use 'none'
        message: >
         Please describe the goals (why) and changes (what) of the PR.
       # message is is optional, a default message is used when not specified.
     must_exclude:
        regex: 'DO NOT MERGE'
-       regex_flag: '' # Optional. Specify the flag for Regex. default is 'i', to disable default use ''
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -6,7 +6,7 @@
   "license": "AGPL-3.0-only",
   "repository": "https://github.com/mergeability/mergeable.git",
   "scripts": {
-    "dev": "LOG_LEVEL=trace NODE_ENV=development nodemon --exec 'npm start'",
+    "dev": "LOG_LEVEL=info NODE_ENV=development nodemon --exec 'npm start'",
```
jusx, line 9: Is this intentional?

docs/README.md
```diff
@@ -231,11 +231,13 @@ completely deleted in the PR.
        message: 'Custom message...' # this is optional, a default message is used when not specified.
     must_include:
        regex: '### Goals|### Changes'
+       regex_flag: '' # Optional. Specify the flag for Regex. default is 'i', to disable default use ''
```
jusx, line 234: I would personally prefer to specify `none` over `''` to override the default. I feel it is more intuitive. Thoughts?

shine2lay: hmm... I agree. putting `''` felt a little off to me as well.

package.json
shine2lay: actually, i think we should remove the LOG_LEVEL, it hasn't been helpful at all and we can individually set it if we need it. Done

PR #275, opened 2020-04-05
feat: add more validations to configuration

### LINKED ISSUE #263: Error not bubbling when 'validate' is not available for each 'when'
### Context
When someone doesn't provide 'validate' field or have a typo, the following error appears in the logs but the error is not propagated to the user 
image

shine2lay:
### Context
Add more checks in validating the config file 
added checks 
- `mergeable` node must not be empty
- for v2: check that `mergeable` is an Array
- for v2: each element in `mergeable` array has both `when` and `validate` keyword
- for v2: `validate` keyword must be non-empty and an array

image

closes #263

### DIFF
--- a/__tests__/unit/configuration/configuration.test.js
+++ b/__tests__/unit/configuration/configuration.test.js
@@ -23,6 +23,7 @@ describe('Loading bad configuration', () => {
     let config = new Configuration(`
     version: not a number
     mergeable:
+      pull_request:
     `)
     expect(config.errors.size).toBe(1)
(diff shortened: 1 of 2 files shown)


jusx:
Nice to have: log.info when there is an error in a config. This is so that we will know how often this happens to understand if there's a better way to improve the ux or the documentation.

shine2lay:
> Nice to have: log.info when there is an error in a config. This is so that we will know how often this happens to understand if there's a better way to improve the ux or the documentation.

We do have this, there is a `log_type` called `config_invalid_yml` for this very purpose. That reminds me i think i need fix that logging as errors are not display as expected

PR #277, opened 2020-04-08
feat: commit validator

### LINKED ISSUE #241: [NEW Validators] check git commit message
In Apache-Airflow dev workflow, we find out only check Github PR title or PR describe is not enough, sometime we should check git commit message to make git history consistency. Otherwise, the git history will look like below

ref: https://github.com/apache/airflow/pull/7144#issuecomment-573490229

shine2lay:
### Context
Added a `commit` validator that runs validations against git commits
sample config
```
- do: commit
  messsage:
    regex: '^(feat|docs|chore|fix|refactor|test|style|perf)(\(\w+\))?:.+$'
    message: 'Custom message' # Semantic release conventions must be followed
    skip_merge: true #Optional, Default is true. Will skip commit with message that includes 'Merge'
    oldest_only: false #Optional, Default is false. Only check the regex against the oldest commit
    single_commint_only: false #Optional, Default is false. only process this validator if there is one commit
```
closes #241

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -404,6 +404,11 @@ Supported events:
     single_commint_only: false #Optional, Default is false. only process this validator if there is one commit
     
 ```
+Supported events:
+
+```js
+'pull_request.*', 'pull_request_review.*'
+```
 
 ### Advanced Logic
 Validators can be grouped together with `AND` and `OR` operators:
--- a/docs/README.md
+++ b/docs/README.md
@@ -399,9 +399,9 @@ Supported events:
   messsage:
     regex: '^(feat|docs|chore|fix|refactor|test|style|perf)(\(\w+\))?:.+$'
     message: 'Custom message' # Semantic release conventions must be followed
-    skip_merge: true #Optional, Default is true. Will skip commit with message that includes 'Merge'
-    oldest_only: false #Optional, Default is false. Only check the regex against the oldest commit
-    single_commint_only: false #Optional, Default is false. only process this validator if there is one commit
+    skip_merge: true # Optional, Default is true. Will skip commit with message that includes 'Merge'
+    oldest_only: false # Optional, Default is false. Only check the regex against the oldest commit
+    single_commint_only: false # Optional, Default is false. only process this validator if there is one commit
     
 ```
 Supported events:
--- a/docs/README.md
+++ b/docs/README.md
@@ -393,6 +393,18 @@ Supported events:
      # all of the message sub-option is optional
 ```
 
+### commit
+```yml
+- do: commit
+  messsage:
+    regex: '^(feat|docs|chore|fix|refactor|test|style|perf)(\(\w+\))?:.+$'
+    message: 'Custom message' # Semantic release conventions must be followed
+    skip_merge: true #Optional, Default is true. Will skip commit with message that includes 'Merge'
+    oldest_only: false #Optional, Default is false. Only check the regex against the oldest commit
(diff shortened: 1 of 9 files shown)


docs/README.md
```diff
@@ -393,6 +393,23 @@ Supported events:
...
+### commit
+```yml
+- do: commit
+  messsage:
+    regex: '^(feat|docs|chore|fix|refactor|test|style|perf)(\(\w+\))?:.+$'
+    message: 'Custom message' # Semantic release conventions must be followed
+    skip_merge: true #Optional, Default is true. Will skip commit with message that includes 'Merge'
+    oldest_only: false #Optional, Default is false. Only check the regex against the oldest commit
+    single_commint_only: false #Optional, Default is false. only process this validator if there is one commit
+
```
jusx, line 405: minor detail: add space after `#`

lib/validators/commit.js
```diff
@@ -0,0 +1,79 @@
...
+const REGEX_NOT_FOUND_ERROR = `Failed to run the test because 'regex' is not provided for 'message' option. Please check README for more information about configuration`
+const DEFAULT_FAIL_MESSAGE = `Some or all of your commit messages doesn't meet the criteria`
+const DEFAULT_SUCCESS_MESSAGE = `Your commit messages met the specified criteria`
+
+class Commit extends Validator {
+  constructor () {
+    super('dependent')
+    this.supportedEvents = [
+      'pull_request.*',
+      'pull_request_review.*'
```
jusx, line 17: I feel we need to be more specific here and only specify `commit` events.

shine2lay: sure, make sense, changed it to pull_request.synchronize only

docs/README.md
shine2lay: fixed

lib/validators/commit.js
shine2lay: actually, that's not true because if we get receive other pull_request event, this validator will get skipped. I actually tested this and i was right. we need to keep `pull_request.*`

jusx: If there are pull_request_reviews.* events would any of the commit change?
If I changed a label, milestones, description or title would any of the commit change?

Do we need to validate for the above events?

It seems redundant. Should we list on specific events where we know the commits will change? Why should commit validation happen for any other events when the commit sets have not changed?

lib/validators/commit.js
```diff
@@ -0,0 +1,79 @@
...
+
+    if (oldestCommitOnly) {
+      orderedCommits = [orderedCommits[0]]
+    }
+
+    const commitMessages = orderedCommits.map(commit => commit.message)
+
+    const result = mustInclude.process(validatorContext, commitMessages, {
+      must_include: {
+        all: true,
```
jusx, line 64: Looks like a convenience flag for the `AND` operator. Should this be available for all options?

lib/validators/commit.js
shine2lay: think about this scenario, 
- pull_request.opened, we check that the commit validation is failing, 
- now i add a label, and if we dont' check `commit` for that event, the same failing validation is now skipped, thus making it `pass`. I think your scenario would only work if we change store the state and change it's status but we are not doing that. we do a fresh check everytime.

lib/validators/commit.js
shine2lay: yes, it is. I think it make sense to have it for ones that deals with an array. i'll create an issue for this.

PR #284, opened 2020-04-12
docs: add a note in the readMe about utilizing anchor for reusability

### LINKED ISSUE #281: avoid duplication in mergeable config
Currently, each `validators` and `actions` have their own sub options and there is no way to generalize them even if the sub options are the same.

For example: 
```
    validate:
      - do: title
        must_include:
          regex: '^(build|chore|ci|docs|feat|fix|improvement|perf|refactor|revert|style|test)(\([a-z0-9_-]+\))?: [^\s].+$'
          regex_flag: 'none'
      - do: commit
        messsage:
          regex: '^(build|chore|ci|docs|feat|fix|improvement|perf|refactor|revert|style|test)(\([a-z0-9_-]+\))?: [^\s].+$'
          regex_flag: 'none'
```
It would be nice to have some way to avoid duplication.
For more context, see #241

shine2lay:
### Context
Add a note in readMe about utilizing yaml anchor references to reduce duplication
 
closes #281

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -488,6 +488,17 @@ similarly, `AND` and `OR` can also be nested at the validator level:
       message: 'If no test plan is necessary, please include test plan: no label'
 ```
 
+To reuse certain parts of the config, you can utilize anchor points that `yaml` provides, like this 
+
+```yml
+  - do: title
+    must_include: &regex_semantic
+      regex: '^(build|chore|ci|docs|feat|fix|improvement|perf|refactor|revert|style|test)(\([a-z0-9_-]+\))?: [^\s].+$'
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -488,6 +488,17 @@ similarly, `AND` and `OR` can also be nested at the validator level:
       message: 'If no test plan is necessary, please include test plan: no label'
 ```
 
+To reuse certain parts of the config, you can utilize anchor points that `yaml` provides, like this
```
jusx, line 491: Would be nice to link to yaml doc here.

shine2lay: done

PR #287, opened 2020-04-15
docs: fix typo in the commit check example

laughedelic:
Copy-pasted this example, got this error:

> Error: Failed to run the 'commit' validator because 'message' option is not found. Please check README for more information about configuration

PR #288, opened 2020-04-15
docs: fix another typo in the commit check example

laughedelic:
I'm on a roll today! 😂 A follow up for #287

This one was worse because it it didn't fail the config validation, just failed the `commit` check itself and  I couldn't figure out why it's failing.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -409,7 +409,7 @@ Supported events:
     message: 'Custom message' # Semantic release conventions must be followed
     skip_merge: true # Optional, Default is true. Will skip commit with message that includes 'Merge'
     oldest_only: false # Optional, Default is false. Only check the regex against the oldest commit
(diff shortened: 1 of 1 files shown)


shine2lay:
@laughedelic whoa, Nice! This is awesome! I guess because it's an optional field, it's just skips that option.

laughedelic:
Yes, it would be nice to have stricter validation of the config, to report "unused" fields?

shine2lay:
@laughedelic yes, i agree, i'll create an issue for that. Thanks again for the fix

laughedelic:
Thanks! Also, is there a way to validate the config locally or remotely without committing it and trying out?

shine2lay:
@laughedelic unfortunately no, you best you can do now is to create a PR with the new `mergeable.yml` config. Mergeable will read the new config from the PR instead of old config from the base branch.

laughedelic:
Ok, thanks!

Issue #244, opened 2020-01-27
Ability to share a config across multiple repos


IllegalCactus:
Hi,

We use Mergeable across a large number of repositories in my organisation. I was wondering if there's a way we could make this workflow a bit simpler by having a 'default' config file across many repos.

The idea I had was: have a convention where you can specify a default `mergeable.yml` either:

- Inside a `.github` repository in your organisation, which is already used to specify org-wide code of conduct and pull request templates: https://help.github.com/en/github/building-a-strong-community/creating-a-default-community-health-file
- Inside a `.mergeable` repository (same as above, but specific to Mergeable)

Mergeable would then use the following priority to decide which file to use

1. A `mergeable.yml` that is added/updated by the PR.
2. Else, a `mergeable.yml` inside the current repo.
3. Else, a `mergeable.yml` inside the 'default' repo.

The use case, as I see it, would be:

- Being able to add Mergeable to a new repo more easily (one click vs. needing to add a file).
- Only have to update config in one place if it's shared across multiple repos (all of our repos use the same config).
- Still able to override the default if need be.

The choice of config file would be all-or-nothing (no merging of files) to keep things simple.

Let me know what you think about this. Is that a feature you would be interested in? I'd be happy to help with a PR if so 🙂 .

shine2lay:
@IllegalCactus I like this a lot. I have a couple of question:
1. How is the 'default' repo determined? I am not familiar with this feature in github and wasn't able to do a quick search to find out.
2. What do you think about a case where i have `mergeable.yml` in the 'default' repo but they don't necessarily want to use it for all repo?

IllegalCactus:
@shine2lay 

1. It's not super well documented. The only mention of this I can find is on the page I linked above ^. GitHub will use contributing guidelines and PR templates from a repo called `.github` as a default for your org, so you don't have to put one in every repo. This doesn't seem to be configurable - it will only ever look in a repo called `.github` by convention. My idea was to just piggyback on this convention. It's nice because it mirrors the path Mergeable currently uses in the repo (`.github/mergeable.yml`). Alternatively, Mergeable could have a convention of looking for a `.mergeable` repo instead.

2. What I had in mind was, you would override the default by including a `mergeable.yml` in your repo. The default would only be used if Mergeable couldn't find one in the specific repo. Is that what you meant?

shine2lay:
@IllegalCactus 
[...]

PR #291, opened 2020-04-18
feat: Add support for default config files

IllegalCactus:
Adds support for specifying default config files in a central repository.
This allows for providing default settings for Mergeable that can be
overriden in individual repos. The purpose is to make it easier to manage
Mergeable across a large number of repos.

The behaviour is quite basic, so there is no support for overriding individual
settings in the configs. Mergeable will use either the default file or the
file in the repo/PR, but never both.

The approach is based on the Probot library which when loading the
config file will look in a repo named '.github' as a fallback. Because
Mergeable can only be granted permission to a specific file path, the
default file needs a somewhat confusing path
('<org>/.github/.github/mergeable.yml').

Addresses:
https://github.com/mergeability/mergeable/issues/244

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -15,3 +15,6 @@ WEBHOOK_SECRET=
 
 # Set this to the webhook URL created with `smee`.
 WEBHOOK_PROXY_URL=
+
+# Needed only for GitHub Enterprise deployments.
+GHE_HOST=
--- a/__tests__/unit/flex.test.js
+++ b/__tests__/unit/flex.test.js
@@ -6,6 +6,7 @@ describe('Test processBeforeValidate and processAfterValidate invocations', asyn
   let context
   let registry = { validators: new Map(), actions: new Map() }
   let action
+
   let config = `
     version: 2
     mergeable:
@@ -28,7 +29,7 @@ describe('Test processBeforeValidate and processAfterValidate invocations', asyn
   `
 
   beforeEach(() => {
-    context = Helper.mockContext('title')
+    context = Helper.mockContext()
     Helper.mockConfigWithContext(context, config)
 
     action = new Action()
@@ -75,12 +76,16 @@ describe('Test processBeforeValidate and processAfterValidate invocations', asyn
 
 describe('#executor', () => {
   test('Bad YML', async () => {
-    let context = Helper.mockContext('title')
+    let context = Helper.mockContext()
+    context.event = 'pull_request'
+    context.payload.action = 'opened'
     Helper.mockConfigWithContext(context, `
       version: 2
-      mergeable:
+        mergeable:
     when: pull_request.*
-    `)
+      `,
+      {files: ['.github/mergeable.yml']}
+    )
 
     context.event = 'pull_request'
     context.payload.action = 'opened'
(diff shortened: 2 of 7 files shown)


__tests__/unit/flex.test.js
```diff
@@ -19,16 +21,19 @@ describe('Test processBeforeValidate and processAfterValidate invocations', asyn
...
-  let configWithMultiple = config + `
+  let configWithMultipleString = configString + `
       - when: pull_request_review.submitted
         validate:
           - do: milestone
             no_empty:
               enabled: true
   `
 
+  let config = yaml.safeLoad(configString)
```
IllegalCactus, line 32: I'm using `safeLoad` across the tests just to maintain the YML here for readability. Let me know if you'd prefer something else, e.g. just putting the objects here directly.

IllegalCactus:
Tested this locally on my machine and it worked OK, but obviously some extra scrutiny would be great. Not sure what your workflow is here.

.env.template
```diff
@@ -15,3 +15,6 @@ WEBHOOK_SECRET=
 
 # Set this to the webhook URL created with `smee`.
 WEBHOOK_PROXY_URL=
+
+# Needed only for GitHub Enterprise deployments.
+GHE_HOST=
```
jusx, line 20: Thanks for adding this.

jusx, line 20: Thanks for adding this.

__tests__/unit/flex.test.js
jusx: This is a good proposal and allows for us to easily manipulate the config in a test and re-mock. However, I'm not seeing this need at the moment in the current tests where `Helper. mockConfigWithContext` is used.

Any reason for wanting this as an object assigned to the variable `config` within the tests?

At the moment I'm deducing that you want it to be obvious that it is `YML`.

IllegalCactus: > Any reason for wanting this as an object assigned to the variable config within the tests?

Hmm, great point. I updated `mockConfigWithContext` to mock `probotContext.config` which returns an object, so that's why. But now that you mentioned it I noticed this would be much simpler if I just flipped it around and made `mockConfigWithContext` continue taking a string, and just do the `yaml.safeLoad` once inside the `probotContext.config` mock.

That way we can continue using strings in the tests and avoid a lot of these changes.

Have updated this now. Let me know what you think :slightly_smiling_face:

IllegalCactus:
Summary of the change incase it helps review (since most of the changes are in tests):

`Configuration.fetchConfigFile` now only loads files directly when the event is a PR that modifies mergeable.yml. In all other cases it uses `probotContext.config` which does the parsing for you and returns an object. In either case all of the parsing happens inside `fetchConfigFile`, which returns an object instead of a YML string.

Hence all the changes to tests.

shine2lay:
@IllegalCactus @jusx tested this feature and ran some smoke testing. Everything works as it should.

kceb:
@IllegalCactus Thank you so much for this feature. It seems like it doesn't support multiple checks currently. Do you know if I'm doing something incorrectly? If I use the same config in my repo, it runs all three checks: https://github.com/mergeability/mergeable/issues/293

shine2lay:
@kceb hmm.. interesting, can you share a link to a PR perhaps? There shouldn't be any different between config from repo and org, they are processed through the same code.

PR #292, opened 2020-04-23
feat: readthedoc documentation + move current docs into old/ folder

shine2lay:
### Context
When we add more features, it'll be harder for new comers to go through the a huge readMe file. So, we should have a documentation site that allow for much easier navigation. 
This PR copied over the current docs into `readthedoc` framework while moving the current ones into `docs/old` folder


jusx:
@shine2lay This PR seem to be removing all notion of the GH README altogether. Please correct me if I'm wrong. I highly recommend moving the README and it's corresponding `md` files to the root folder. 

We need to at least have a README on GH. We can point to readthedocs for the main usage and configuration documentation -- which is a majority of what's in the README.

shine2lay:
@jusx I moved all the current readMe related files into `docs/old` folder

jusx:
@shine2lay I understand that you have it in /docs/old -- however, GH is not going to process that as a README for the repo. As I understand it, by merging this PR as is, this repos will be without any README and I don't think we should do that. I'd prefer to retain the old README and have time to fix it appropriately and point certain parts to readthedocs appropriately.

shine2lay:
@jusx got it, addressed

PR #297, opened 2020-05-05
feat: ability to create multiple checks

### LINKED ISSUE #225: Allow to split conditions different validators in separate checks
One GitHub app can create multiple checks, so it would be nice if mergeable would support this (unless it does and I just don't know how set it up then I'll be thankful for any tips)

image

shine2lay:
### Context
each recipe now has the ability to add `name` field, which will show up when the checks are created. s
example config:

```
version: 2
mergeable:
    name: 'JIRA ticket in Title'
    validate:
      - do: title
        must_include:
          regex: '\[[a-zA-Z]{2,6}-\d{1,7}\]'
          regex_flag: 'none'
          message: >
            Please include ticket in title in square brackets (ex: [AAA-123])
    name: 'Non Empty Description'
    validate:
      - do: description
        no_empty:
          enabled: true
```

Note: if no name are provided, it'll work as is currently and only create a check for the last `recipe`

output :
image

closes #225


lib/flex.js
```diff
@@ -181,7 +181,7 @@ const processPreActions = async (context, registry, config) => {
     if (isEventInContext(rule.when, context)) {
       registry.actions.forEach(action => {
         if (action.isEventSupported(`${context.event}.${context.payload.action}`)) {
-          promises.push(action.processBeforeValidate(context))
+          promises.push(action.processBeforeValidate(context, rule, rule.name))
```
jusx, line 184: Why not add rule and rule.name to `context`? That's what it's for so that each function has the context it needs and such that we aren't constantly adding new args.

shine2lay: well, context is not individual to each `recipe`. if there are multiple `recipe`, context will have to hold that information separate from the settings

jusx: Got it. Since you're passing in the rule and name is a property of rule, can't the function obtain `name` from within? Any reason to have the `rule` object and `name` property of rule as two separate args?

shine2lay: for `beforeValidate` yes, we can grab the name from `rule` since it doesn't require any specific settings to process beside to log it. in `aftervalidate` we only pass in the relevant settings which is based on whether the validations were a `pass` or `fail`. So we need a `name` field in aftervalidate for sure, i also added in for `beforevalidate` for symmetry.

PR #299, opened 2020-05-12
feat: allow approvals to be limited to team members

### LINKED ISSUE #236: Review required from members of a Team
Hi,
we have the following situation:
We got two teams with write access to our repository and we'd like to have reviews/approvals of 2 members of the first team and 1 member of the second team to approve a pull-request to be mergeable. We tried some different approaches with mergeable: Insert the Team-Names as reviewers, Insert the team names in the CodeOwners file, but both didn't work (in the second scearnio mergeable didn't finish at all, it was running for almost half an hour).
Do you have any idea how our requirement could be solved using mergeable? Or do you maybe plan to implement to support team-names as required reviewers?

Thanks you very much!

shine2lay:
Add the ability to limit the approval validator to team members
s
sample config:
```
- do: approvals
      min:
        count: 2 # Number of minimum reviewers. In this case 2.
        message: 'Custom message...'
      limit:
        teams: ['org/team_slug']  # when the option is present, only the approvals from the team members will count
```

closes #236

### DIFF
--- a/lib/validators/approvals.js
+++ b/lib/validators/approvals.js
@@ -4,6 +4,8 @@ const Assignees = require('./options_processor/assignees')
 const RequestedReviewers = require('./options_processor/requestedReviewers')
 const consolidateResult = require('./options_processor/options/lib/consolidateResults')
 const constructOutput = require('./options_processor/options/lib/constructOutput')
+const constructErrorOutput = require('./options_processor/options/lib/constructErrorOutput')
+const TeamNotFoundError = require('../errors/teamNotFoundError')
 const options = require('./options_processor/options')
 const _ = require('lodash')
 
@@ -40,6 +42,12 @@ class Approvals extends Validator {
       delete validationSettings.block
     }
 
+    let limitOption = null
+    if (validationSettings.limit) {
+      limitOption = validationSettings.limit
+      delete validationSettings.limit
+    }
+
     let reviews = await context.github.paginate(
       context.github.pulls.listReviews.endpoint.merge(
         context.repo({ pull_number: this.getPayload(context).number })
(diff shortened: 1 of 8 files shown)


lib/validators/approvals.js
```diff
@@ -34,98 +36,164 @@ class Approvals extends Validator {
   }
 
   async validate (context, validationSettings) {
+    // const res = await getTeamMembers(context, 'test-mergeable', 'reviewers')
```
jusx, line 39: Is this intentional? If so, I recommend deleting this line all together.

jsedlace:
@jusx Hello Justin, do you have some plan when this functionality will be deployed and can be used if I have mergeable configured through github.com?

shine2lay:
@jsedlace We are waiting until we got a few more features in before releasing all of it together, we expect this feature to be available around Mid - End of July. you should get an email if you have installed mergeable because we'll need to require more app permission for this feature to be available.

jsedlace:
@shine2lay Thanks for fast reply. In this case we will see request for more permissions, in other cases maybe not - how often you release and how it is tracked in the repo? Tags/branches/github release?

shine2lay:
@jsedlace ~ We do continuous deployment whenever the code is merged into master and the tests are passing. All the codes that need more app permission simply throw errors when attempting to use them. We currently doesn't have a versioning but we recently started a [CHANGELOG](https://mergeable.readthedocs.io/en/latest/changelog.html) to track new feature that has already been deployed.

jusx:
@jsedlace Most of our deployment so far does not require permission changes. We are starting to work on features that requires permissions not formerly configured. Our goal is to minimize the amount of GH permission authorization by making these permission changes at the same time. The target date for this is by end of July.

Since we are auto-deploy on merge to `master` we have stopped creating github releases. Would you like us to notify you when we update the github app permission settings?

PR #300, opened 2020-05-13
fix: after validate

shine2lay:
### Context
In the #297 , I accidentally left out parameter changes to afterValidate.
This had a small impact on `comment` action but, it is very low impact.


shine2lay:
merging because this is impacting production and it's minor fix

PR #301, opened 2020-05-15
feat: delete old comments before creating new one

### LINKED ISSUE #157: Automatically remove obsolete comments by mergeable
When mergeable comments on an issue because a validation fails it adds all the information why. That's helpful. But when the issue is fixed, e.g. by adding a label and a description, the comment stays. That can be quite confusing and it also adds some noise to the issue.

It would be great if mergeable would remove its own comment if it doesn't apply anymore.

This would also prevent multiple mergeable comments in a row, when an issue is edited and mergeable is triggered by each part of the edit.

Example: Add an issue without description and label, mergeable comments, add a label, mergeable comments again, add a description, no further comment, but the two comments which are there now, don't apply anymore.

shine2lay:
Old comments are removed by default before creating a new one on an issue

new config : 
```
    - do: comment
      payload:
        body: >
          Your very long comment can go here.
      old_comment: 'delete' # delete by default, if you wish to leave the old comments use 'leave'
```

closes #157

### DIFF
--- a/docs/actions/comment.rst
+++ b/docs/actions/comment.rst
@@ -7,6 +7,7 @@ Comment
       payload:
         body: >
           Your very long comment can go here.
+      old_comment: 'delete' # delete by default, if you wish to leave the old comments use 'leave'
 
 Supported Events:
 ::
--- a/docs/actions/comment.rst
+++ b/docs/actions/comment.rst
@@ -7,7 +7,7 @@ Comment
       payload:
         body: >
           Your very long comment can go here.
-      old_comment: 'delete' # delete by default, if you wish to leave the old comments use 'leave'
+      leave_old_comment: true # Optional, by default old comments are deleted, if true, old comments will be left alone
 
 Supported Events:
 ::
--- a/lib/actions/comment.js
+++ b/lib/actions/comment.js
@@ -12,7 +12,7 @@ const fetchCommentsByMergeable = async (context, issueNumber) => {
     context.repo({ issue_number: issueNumber })
   )
 
-  const botName = process.env.BOT_NAME ? process.env.BOT_NAME : 'Mergeable'
+  const botName = process.env.APP_NAME ? process.env.APP_NAME : 'Mergeable'
 
   return comments.data.filter(comment => comment.user.login === `${botName}[bot]`)
 }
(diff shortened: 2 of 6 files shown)


docs/actions/comment.rst
```diff
@@ -7,6 +7,7 @@ Comment
       payload:
         body: >
           Your very long comment can go here.
+      old_comment: 'delete' # delete by default, if you wish to leave the old comments use 'leave'
```
jusx, line 10: This looks more like a boolean. do you envision there be more options here? And if so would it make more sense for those to be options?

lib/actions/comment.js
```diff
@@ -7,6 +7,27 @@ const createComment = async (context, issueNumber, body) => {
...
+  const botName = process.env.BOT_NAME ? process.env.BOT_NAME : 'Mergeable'
+
+  return comments.data.filter(comment => comment.user.login === `${botName}[bot]`)
+}
+
+const deleteOldComments = async (context, issueNumber) => {
+  const oldComments = await fetchCommentsByMergeable(context, issueNumber)
+
+  for (let comment of oldComments) {
+    console.log(comment)
```
jusx, line 24: Probably unintentional and should be removed.

lib/actions/comment.js
```diff
@@ -7,6 +7,27 @@ const createComment = async (context, issueNumber, body) => {
   )
 }
 
+const fetchCommentsByMergeable = async (context, issueNumber) => {
+  const comments = await context.github.issues.listComments(
+    context.repo({ issue_number: issueNumber })
+  )
+
+  const botName = process.env.BOT_NAME ? process.env.BOT_NAME : 'Mergeable'
```
jusx, line 15: May want to add a note to the README for installation, etc.

docs/actions/comment.rst
shine2lay: I imagine, there could be either `edit` or `overwrite` option (not decided on the naming yet), similar to how code-cov does it, it maybe helpful to enforce rules on issues without too much noise.

shine2lay: I see your point, i'll change it into a boolean

PR #302, opened 2020-05-17
feat: merge action

### LINKED ISSUE #201: Merge on green label
**Context**

Waiting on status checks or an author to respond to a PR approval can increase the time to merge.  One solution is automating the merge process using labels or PR comments.

Pseudo label example:
```yml
- do: merge
  labels: 'merge on green', 'merge when passing'
  type: squash #options: merge, squash, rebase
  delete_branch: true # deletes the branch once merged.
  delay: 30 # Wait 30 sec. before performing merge.
```

or, using the comment approach:
```
@mergable-bot merge on green
```

Other bot examples:
- [Ranger](https://github.com/marketplace/ranger)
- [Mergify](https://github.com/marketplace/mergify)

shine2lay:
Context
Adds merge action 

example config 
```
    - do: merge
      merge_method: 'merge' # Optional , default is 'merge'. Other options : 'rebase', 'squash'
```
This feature does require more app permission so it won't work right now for mergeable

closes #201

### DIFF
--- a/lib/actions/merge.js
+++ b/lib/actions/merge.js
@@ -1,0 +1,52 @@
+const { Action } = require('./action')
+
+const checkIfMerged = async (context, prNumber) => {
+  let status = true
+
+  // return can be 204 or 404 only
+  try {
+    await context.github.pulls.checkIfMerged(
+      context.repo({ pull_number: prNumber })
+    )
+  } catch (err) {
+    if (err.status === 404) {
+      status = false
+    } else {
+      throw err
+    }
+  }
+
+  return status
+}
+
+class Comment extends Action {
+  constructor () {
+    super('merge')
+    this.supportedEvents = [
+      'pull_request.*'
+    ]
+  }
+
+  // there is nothing to do
+  async beforeValidate () {}
+
+  async afterValidate (context, settings, results) {
+    const prNumber = this.getPayload(context).number
+    const isMerged = await checkIfMerged(context, prNumber)
+
+    if (!isMerged) {
+      let mergeMethod = settings.merge_method ? settings.merge_method : 'merge'
+      try {
+        await context.github.pulls.merge(context.repo({ pull_number: prNumber, merge_method: mergeMethod }))
+      } catch (err) {
+        // skip on known errors
+        // 405 === Method not allowed , 409 === Conflict
+        if (!(err.status === 405 || err.status === 409)) {
+          throw err
+        }
+      }
+    }
+  }
+}
+
+module.exports = Comment
--- a/lib/actions/merge.js
+++ b/lib/actions/merge.js
@@ -1,4 +1,7 @@
 const { Action } = require('./action')
+const UnSupportedSettingError = require('../errors/unSupportedSettingError')
+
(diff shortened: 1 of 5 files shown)


lib/actions/merge.js
```diff
@@ -0,0 +1,52 @@
...
+      status = false
+    } else {
+      throw err
+    }
+  }
+
+  return status
+}
+
+class Comment extends Action {
```
jusx, line 22: You mean `Merge` isn't it?

lib/actions/merge.js
```diff
@@ -0,0 +1,52 @@
...
+
+  // there is nothing to do
+  async beforeValidate () {}
+
+  async afterValidate (context, settings, results) {
+    const prNumber = this.getPayload(context).number
+    const isMerged = await checkIfMerged(context, prNumber)
+
+    if (!isMerged) {
+      let mergeMethod = settings.merge_method ? settings.merge_method : 'merge'
```
jusx, line 38: Should we validate valid mergeMethod? Return a warning when loading config with wrong mergeMethods?

lib/actions/merge.js
```diff
@@ -0,0 +1,52 @@
...
+      try {
+        await context.github.pulls.merge(context.repo({ pull_number: prNumber, merge_method: mergeMethod }))
+      } catch (err) {
+        // skip on known errors
+        // 405 === Method not allowed , 409 === Conflict
+        if (!(err.status === 405 || err.status === 409)) {
+          throw err
```
jusx, line 45: Can't remember the behavior now, but what is the user experience when this happens? Should we execute a comment to let the user know about failures?

lib/actions/merge.js
```diff
@@ -0,0 +1,52 @@
...
+        }
+      }
+    }
+  }
+}
+
+module.exports = Comment
```
jusx, line 52: Same as above. Did you mean for this to be `Merge` ?

lib/actions/merge.js
shine2lay: We currently have no feature to give feedback when any `action` throw error. right now. even in validators, we only give feedback when it's a `pull_request` by creating a check.

jusx: I highly recommend putting more thought around a user feedback mechanism now that we are starting to support advanced actions. 

IMO For actions that does not support a feedback mechanism i.e. `check` we use comments in within the PR/Issue.

shine2lay: yes, i've been thinking about it. I think it should go into a separate PR that handles all `action` errors. since that scope is bigger than just one action. I'll fix PR this so that it will throw proper error

jusx: Agreed.

lib/actions/merge.js
jusx: Any thoughts on this?

shine2lay: fixed

lib/actions/merge.js
shine2lay: fixed

kevnm67:
@shine2lay looking forward to trying this out!  Thanks for entertaining and implementing this feature 🥇

PR #304, opened 2020-05-21
feat: limit option for stale validator

### LINKED ISSUE #221: Stale PR comments only on work days feature?
The feature of commenting stale pull requests is nice but it not necessarily makes any good when these comments are posted on weekends when all developers are at home. They often ignore email notifications especially if they use their private GitHub accounts at work. And I can’t blame them. It’s their time off. 

Maybe it’s not the most important feature request but it would be nice to be able to set the days when stale PR comments are posted (or the days when they’re not supposed to be posted). 

I’m curious if other people would find it useful. Let me know by thumbs up or thumbs down reaction.

shine2lay:
Context
allow stale to only run if certain conditions are met
```
    - do: stale
      days: 20 # number of days ago.
      type: pull_request, issues # what items to search for.
      limit: # Optional, run only if the specific options are met
        time_zone: 'America/Los_Angeles' # Optional, UTC time by default, for valid timezones see `here <https://momentjs.com/timezone/>`_
        hours_between: ['9', '17'] # Optional, 24 hours by default, run only if [0] >= Hour Now <= [1]
        days_of_Week: ['Mon', 'Tue', 'Wed', 'Thu', 'Fri'] # Optional, 7 days a week by default, specific the days of the week in which to run the validator
```

closes #221

### DIFF
--- a/docs/validators/stale.rst
+++ b/docs/validators/stale.rst
@@ -6,7 +6,7 @@ Stale
     - do: stale
       days: 20 # number of days ago.
       type: pull_request, issues # what items to search for.
-      limit: # Optional, run only if the specific options are met
+      time_constraint: # Optional, run the validator only if it in within the time constraint
         time_zone: 'America/Los_Angeles' # Optional, UTC time by default, for valid timezones see `here <https://momentjs.com/timezone/>`_
         hours_between: ['9', '17'] # Optional, 24 hours by default, run only if [0] >= Hour Now <= [1]
         days_of_Week: ['Mon', 'Tue', 'Wed', 'Thu', 'Fri'] # Optional, 7 days a week by default, specific the days of the week in which to run the validator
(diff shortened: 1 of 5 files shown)


docs/validators/stale.rst
```diff
@@ -4,13 +4,17 @@ Stale
 ::
 
     - do: stale
-    days: 20 # number of days ago.
-    type: pull_request, issues # what items to search for.
+      days: 20 # number of days ago.
+      type: pull_request, issues # what items to search for.
+      limit: # Optional, run only if the specific options are met
```
jusx, line 9: TOL: This configuration type might be something we want add to another validators and actions in the future.

shine2lay: hmm.. in that case, maybe we should call it `time_limit` so it can be reusable in all actions and validators. i'll fix it

shine2lay: to close the loop here, i ended up name it `time_constraint` instead

Issue #233, opened 2019-10-22
Check never finishes


tiferrei:
Hi my check seems to never finish if I add comment actions.
Here's my config:
```yaml
version: 2
mergeable:
  - when: pull_request.*, pull_request_review.*
    validate:
      - do: title
        must_exclude:
          regex: ^WIP
      - do: label
        must_exclude:
          regex: WIP|wip|work-in-progress
        must_include:
          regex: deploy|DEPLOY
    fail:
      - do: comment
        payload:
          body: >
            **PR Validation Failed**: DEPLOY has not been acknowledge or PR in WIP. Please add the DEPLOY label and remove WIPs.
    pass:
      - do: comment
        payload:
          body: >
            **PR Validation Passed**: DEPLOY acknowledge. Can now merge.
```
It gets to comment the pass/fails, but the check never finishes running.

shine2lay:
@tiferrei This is a known issue, if you specify either `pass` or `fail`, you also need to pass in the defaults as they get overwritten

default checks for `pass` :
```
- do: checks,
  status: 'success' # Can be: success, failure, neutral, cancelled, timed_out, or action_required
  payload:
    title: 'Mergeable Run have been Completed!'
    summary: `All the validators have returned 'pass'! \n Here are some stats of the run: \n {{validationCount}} validations were ran`
```

default checks for `fail` : 
```
- do: checks # default fail case
  status: 'failure' # Can be: success, failure, neutral, cancelled, timed_out, or action_required
  payload:
    title: 'Mergeable Run have been Completed!'
    summary: "### Status: {{toUpperCase validationStatus}}
             
                     Here are some stats of the run:
                     {{validationCount}} validations were ran.
                     {{passCount}} PASSED
                     {{failCount}} FAILED
                   "
    text: "{{#each validationSuites}}
          #### {{{statusIcon status}}} Validator: {{toUpperCase name}}
          {{#each validations }} * {{{statusIcon status}}} ***{{{ description }}}***
                 Input : {{{details.input}}}
                 Settings : {{{displaySettings details.settings}}}
                 {{/each}}
          {{/each}}"
```

tiferrei:
Ah thanks, got it. Is this behaviour anywhere in the README?

shine2lay:
@tiferrei I don't think it exist in README yet, i just took a brief look and couldn't find it.

colinking:
[...]

PR #305, opened 2020-05-22
feat: fill in missing param for `checks` by using default values

shine2lay:
Context

When user provide partial parameter list for checks option, fill in the rest using default values

for more context, see https://github.com/mergeability/mergeable/issues/233#issuecomment-632211789

### DIFF
--- a/__tests__/unit/configuration/transformers/v2Config.test.js
+++ b/__tests__/unit/configuration/transformers/v2Config.test.js
@@ -110,7 +110,7 @@ test('only pass, fail defaults ignore recipes that are not for pull_requests', (
   expect(transformed.mergeable[0].error).toEqual([])
 })
 
-test('default checks fill in missing fields', () => {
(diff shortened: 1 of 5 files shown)


shine2lay:
I took that comment to mean use default if field is missing in general for `checks`. i.e if they don't provide `summary` field, use default also as well. Do you think we should leave it at `title` only?

jusx:
> I took that comment to mean use default if field is missing in general for `checks`. i.e if they don't provide `summary` field, use default also as well. Do you think we should leave it at `title` only?

The original issue and use case was around lack of feedback when a required property is not specified and the PR check hanging. For the check action, I suggest we:

1. default any GH APi required fields when not specified.
2. Update the docs about this behavior.

shine2lay:
Ok, i'll fix it

shine2lay:
@jusx  PTAL

PR #306, opened 2020-05-22
feat: assign PR/Issue to author

hemeroc:
**Usecase:** As a user I want to be able to automatically assign a PR/Issue not only to a predefined person but also to the author of the PR/Issue.

**Solution:** Besides allowing a fixed list of assignees one should be able to put in a placeholder like `@author` to assign the PR/Issue to the creator

**Sample usage:**
```yaml
version: 2
mergeable:
  - when: pull_request.*, issues.*
    name: 'Auto assign to author if no assignee is set'
    validate:
      - do: assignee
        min:
          count: 1
    fail:
      - do: assign
        assignees: [ '@author' ]
```

**Remark:** I have no hard feelings about the placeholder name, I just used `@author` because the `@` is never part of a github username

### DIFF
--- a/lib/actions/assign.js
+++ b/lib/actions/assign.js
@@ -20,6 +20,13 @@ const checkAssignee = async (context, issueNumber, assignee) => {
   return isValidAssignee(checkResult) ? assignee : null
 }
 
+const mapSpecialAssignee = async (payload, assignee) => {
+  if (assignee.toLowerCase() === '@author') {
+    return payload.user.login
+  }
+  return assignee
+}
+
 class Assign extends Action {
   constructor () {
     super('assign')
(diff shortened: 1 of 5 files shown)


lib/actions/assign.js
```diff
@@ -20,6 +20,13 @@ const checkAssignee = async (context, issueNumber, assignee) => {
   return isValidAssignee(checkResult) ? assignee : null
 }
 
+const mapSpecialAssignee = async (payload, assignee) => {
```
shine2lay, line 23: just a nitpick, does this function needs to be `async`?

lib/actions/assign.js
```diff
@@ -35,7 +42,7 @@ class Assign extends Action {
   async afterValidate (context, settings, name, results) {
     const payload = this.getPayload(context)
     const issueNumber = payload.number
-    const assignees = settings.assignees
+    const assignees = await Promise.all(settings.assignees.map(assignee => mapSpecialAssignee(payload, assignee)))
```
shine2lay, line 45: if `mapSpecialAssignee` is not async then you could remove `await Promise.all` and make it simpler

lib/actions/assign.js
hemeroc: You are absolutely right this is not needed anymore.

hemeroc:
@shine2lay I incorporated your feedback and added a line to the changelog.

shine2lay:
@hemeroc Looks great! i'll go ahead and merge it in

PR #307, opened 2020-05-24
feat: validate settings and throw error if unexpected or unknown settings are found

### LINKED ISSUE #289: Throw error if there are un-used options
Context : 
For optional fields, we check whether it exists and skips it if it is undefined. This doesn't account for cases where there are typos for the optional fields.

It would be useful to check if there are un-used fields in validator and throw errors if true.

for more context: see #288

shine2lay:
User Story : As a user, i want to know if there are unknown settings in my config so that i can fix it.

image

closes #289

### DIFF
--- a/lib/validators/validator.js
+++ b/lib/validators/validator.js
@@ -1,6 +1,10 @@
+const _ = require('lodash')
 const EventAware = require('../eventAware')
 const options = require('./options_processor/options')
 const logger = require('../logger')
+const UnSupportedSettingError = require('../errors/unSupportedSettingError')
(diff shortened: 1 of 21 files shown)


lib/validators/validator.js
```diff
@@ -49,6 +69,29 @@ class Validator extends EventAware {
     }
     this.log.info(JSON.stringify(usageLog))
   }
+
+  validateSettings (supportedSettings, settingToCheck, nestings = []) {
+    console.log('toCheck', settingToCheck)
```
jusx, line 74: Do we need this here?

PR #311, opened 2020-05-30
fix: `regex_flag` to supported Settings

shine2lay:
`regex_flag` setting were missing from #307 

closes #307


shine2lay:
merging in w/o approval because it is breaking production and doesn't introduce new logic

PR #315, opened 2020-06-01
fix: commit validator supported Settings

shine2lay:
Context

commit validator name had a typo + supported setting is missing `regex_flag` variable


shine2lay:
merging in without approval because no new logic is added

PR #316, opened 2020-06-01
fix: test supported settings

shine2lay:
Context
fix the tests to make sure the `validateSetting` is called to make sure all the valid settings implemented are a part of `supportedSettings`

this PR includes fixes to the `supportedSettings` that were missing, which is currently breaking production for certain settings


shine2lay:
merging in without approval because production is broken and doesn't introduce new logic

PR #318, opened 2020-06-18
feat: contents validator

### LINKED ISSUE #207: Rules based on the content of files
How can I have a rule that checks the content of the diff ?
For example, fail PRs with/when:
* image tags using http:// instead of https://
* secret keys exists in code

shine2lay:
### Add contents validator
config 
```
    - do: contents
      files: # determine which files contents to validate
        pr_diff: true # If true, validator will grab all the added and modified files in the head of the PR
        ignore: ['.github/mergeable.yml'] # Optional, default ['.github/mergeable.yml'], pattern of files to ignore
      must_include:
         regex: 'yarn.lock'
         message: 'Custom message...'
      must_exclude:
         regex: 'package.json'
         message: 'Custom message...'
      begins_with:
         match: 'A String' # or array of strings
         message: 'Some message...'
      ends_with:
         match: 'A String' # or array of strings
         message: 'Come message...'
```

sample output

image

closes #207

### DIFF
--- a/__tests__/unit/validators/contents.test.js
+++ b/__tests__/unit/validators/contents.test.js
@@ -1,7 +1,7 @@
 const Helper = require('../../../__fixtures__/unit/helper')
 const Contents = require('../../../lib/validators/contents')
 
-test.only('files pr_diff option works correctly', async () => {
+test('files pr_diff option works correctly', async () => {
   const contents = new Contents()
   const settings = {
     do: 'contents',
@@ -21,7 +21,7 @@ test.only('files pr_diff option works correctly', async () => {
   expect(validation.status).toBe('pass')
 })
 
-test.only('files ignore option works correctly', async () => {
+test('files ignore option works correctly', async () => {
   const contents = new Contents()
   let settings = {
     do: 'contents',
--- a/__tests__/unit/validators/contents.test.js
+++ b/__tests__/unit/validators/contents.test.js
@@ -41,14 +41,14 @@ test.only('files ignore option works correctly', async () => {
     do: 'contents',
     files: {
       pr_diff: true,
-      ignore: []
+      ignore: ['package.json']
     },
     must_exclude: {
       regex: 'test'
     }
   }
 
-  validation = await contents.processValidate(createMockContext(['.github/mergeable.yml', 'package.json'], 'string'), settings)
+  validation = await contents.processValidate(createMockContext(['package.json'], 'string'), settings)
   expect(validation.status).toBe('pass')
 })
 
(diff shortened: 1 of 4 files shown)


__tests__/unit/validators/contents.test.js
```diff
@@ -0,0 +1,65 @@
+const Helper = require('../../../__fixtures__/unit/helper')
+const Contents = require('../../../lib/validators/contents')
+
+test.only('files pr_diff option works correctly', async () => {
```
jusx, line 4: Is the `only` intentional?

__tests__/unit/validators/contents.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+  let validation = await contents.processValidate(createMockContext(['package.json'], 'testString'), settings)
+  expect(validation.status).toBe('fail')
+  expect(validation.validations[0].description.includes('package.json')).toBe(true)
+
+  validation = await contents.processValidate(createMockContext(['package.json'], 'string'), settings)
+  expect(validation.status).toBe('pass')
+})
+
+test.only('files ignore option works correctly', async () => {
```
jusx, line 24: Is `only` intentional?

lib/validators/contents.js
```diff
@@ -0,0 +1,96 @@
...
+)
+
+const getContent = async (context, path) => {
+  return context.github.repos.getContents(context.repo({
+    path: path,
+    ref: context.payload.pull_request.head.sha
+  })).then(res => {
+    return Buffer.from(res.data.content, 'base64').toString()
+  }).catch(error => {
+    console.log(error)
```
jusx, line 90: Probably want to remove this debugging statement.

lib/validators/contents.js
```diff
@@ -0,0 +1,96 @@
...
+    let changedFiles = result.data
+      .filter(file => !matchesIgnoredPatterns(file.filename, patternsToIgnore))
+      .filter(file => file.status === 'modified' || file.status === 'added')
+      .map(file => file.filename)
+
+    let failedFiles = []
+    for (let file of changedFiles) {
+      const content = await getContent(context, file)
+
+      const processed = this.processOptions(parseOption, content)
```
jusx, line 64: Will `processOption` work properly when content == null?

shine2lay: if file we are fetching doesn't exist, the `getContent` will throw error. i don't think `null` can exist here

jusx: You are returning null when it is a 404. 

ref: https://github.com/mergeability/mergeable/pull/318/files#diff-4c8888c6c6e37cf61921bb33ec55d224R90

shine2lay: you are absolutely right, i missed that. I'll create a PR to fix that. 
In terms of severity, i think it's low because contents can only work with `pr_diff` ATM, which are files known to exist.

PR #321, opened 2020-07-02
feat: re-run checks if base branch has modified config file

### LINKED ISSUE #153: Checks are not re-run if mergeable.yml has changed
Our use case is the following. We'd like to have our current milestone set in the mergeable.yml so that **only** PRs targeting the current milestone are able to be merged. Once we release this version, we'd like to change the milestone in mergeable.yml and have all PRs that were previously targeting this version re-run, so that the check passes as they now target the current milestone.

I believe this goes beyond this use case though. If you change your configuration on mergeable.yml, all PRs which mergeable reported something should re-run in order to reflect the up-to-date state with this new configuration.

shine2lay:
### Feature

Whenever  the config file in the base branch is add/modified, mergeable will trigger on all PR that is against the base branch using the new configuration

closes #153


shine2lay:
@jusx I think we did that in order to emit events like `pull_request.closed`, i've reverted it

PR #322, opened 2020-07-02
fix: handle case when content is not found

shine2lay:
Address the case in `contents` validator when the content file is not found.

Behavior implemented : 
If content is not found the validator will `fail` and `filename (Not found)` will be added to failed file list

### DIFF
--- a/__tests__/unit/validators/contents.test.js
+++ b/__tests__/unit/validators/contents.test.js
@@ -76,7 +76,6 @@ test('fail gracefully if content is not found', async () => {
   }
 
   let validation = await contents.processValidate(context, settings)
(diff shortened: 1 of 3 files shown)


__tests__/unit/validators/contents.test.js
```diff
@@ -52,6 +52,35 @@ test('files ignore option works correctly', async () => {
...
+    files: {
+      pr_diff: true
+    },
+    must_exclude: {
+      regex: 'test'
+    }
+  }
+
+  let validation = await contents.processValidate(context, settings)
+  console.log(validation.validations)
```
jusx, line 79: please remove this.

PR #326, opened 2020-07-10
fix: The file should be named CODEOWNERS

pkuczynski:
GitHub expects `CODEOWNERS` not `CODEOWNER`, according to their documentation

https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners


shine2lay:
It seems we are fetching the correct file, the typo is only in the documentation

pkuczynski:
Yes, that's what I figure out too...

PR #333, opened 2020-07-11
feat: allow @author in comment action

### LINKED ISSUE #328: Mention in the comment author of the PR/issue
I am wondering if it would be possible to refer to PR/issue author in `do: comment`?

shine2lay:
I also refactor so that the special annotations are reusable 
closes #328


jusx:
This works in most cases but. It can be problematic when the real github user @author is involved. I recommend using a different prefix symbol that is not native to Github.

shine2lay:
@jusx what you say is true, but this annotation has already been implemented for assignee a while back. So for the sake of backward compatibility, I would like to suggest an escape character instead, maybe `\` , infront of the annotation to skip the processing?

shine2lay:
@jusx PTAL

jusx:
@shine2lay I would prefer to change the prefix for assignee but a slash will work -- less preferable imo.

PR #339, opened 2020-07-15
fix: bug where pass, fail or error is undefined

shine2lay:
this bug was introduced in #338


shine2lay:
merging in without approval because this is breaking production

PR #349, opened 2020-08-02
fix: github events link in README.md

kevnm67:
Trivial update.
- Updated the github events link to point to its new home.


shine2lay:
this is no longer relevant as we have moved our documentation to mergeable.readthedocs.io

PR #361, opened 2020-08-16
chore: add merge and changelog to mergeable config

shine2lay:
- Added auto merging when mergeable is passing
- Added check to check if `docs/changelog.rst` is modified when the PR is a `feat or fix`


shine2lay:
Seems PR is not yet ready for review. will fix later

PR #373, opened 2020-09-16
feat: Allow specifying files to match (#5)

samj1912:
Fixes #373

### DIFF
--- a/__tests__/unit/validators/size.test.js
+++ b/__tests__/unit/validators/size.test.js
@@ -274,6 +274,20 @@ describe('PR size validator', () => {
     expect(validation.validations[1].status).toBe('pass')
     expect(validation.validations[2].description).toBe('PR size for total additions + deletions is OK!')
     expect(validation.validations[2].status).toBe('pass')
+
+    // Explicitly setting ignore to blank so
+    // that we can test for matches.
+    settings.match = ['not_*.js']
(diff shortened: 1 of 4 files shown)


samj1912:

__tests__/unit/validators/size.test.js
```diff
@@ -274,6 +274,17 @@ describe('PR size validator', () => {
     expect(validation.validations[1].status).toBe('pass')
     expect(validation.validations[2].description).toBe('PR size for total additions + deletions is OK!')
     expect(validation.validations[2].status).toBe('pass')
+
+    settings.match = ['not_*.js']
```
shine2lay, line 278: I don't think this isolate the `match` settings, could you please a test case where only `match` setting is tested? 
Here, i am not sure if this is working because of `ignore` or `match` since both are present

shine2lay (changes requested):
sorry for the late response, i was on a break. I had one minor comment in the test, everything else looks great! thanks

samj1912:
@shine2lay addressed the comment on tests :)

__tests__/unit/validators/size.test.js
samj1912: Added a line below to set `ignore` to a blank array :)

shine2lay (approved):

PR #378, opened 2020-09-17
fix: Fix the comment action to work correctly with the scheduler

### LINKED ISSUE #376: Stale comments are broken
There are 2 causes to this. 1 delete old comments logic and 2 custom logic to render special annotations in comments. Both use getPayload which returns null since the context.event is 'schedule' which doesn't exist in the context.payload

samj1912:
Fixes #376.


samj1912:

shine2lay (approved):

PR #379, opened 2020-09-17
feat: Add support for labels and close actions for stale

samj1912:
Fixes #379

### DIFF
--- a/lib/actions/close.js
+++ b/lib/actions/close.js
@@ -11,7 +11,8 @@ class Close extends Action {
     super('close')
     this.supportedEvents = [
       'pull_request.*',
-      'issues.*'
+      'issues.*',
+      'schedule.repository'
     ]
   }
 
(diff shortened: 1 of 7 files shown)


samj1912:

lib/actions/close.js
```diff
@@ -11,20 +11,26 @@ class Close extends Action {
...
   async beforeValidate () {}
 
   async afterValidate (context, settings, name, results) {
-    const payload = this.getPayload(context)
-    const issueNumber = payload.number
-
-    return closeIssue(
-      context,
-      issueNumber
+    let scheduleResults = results && results.validationSuites && results.validationSuites[0].schedule
```
shine2lay, line 23: have you tested these with non `schedule.repository` flows? I don't remember if normal flows have `.schedule` properties thus this might fail

samj1912: yes I have. For non schedule flows the items variable below is set to getPayload(context) instead of the schedule results since schedule results will be undefined.

samj1912: This logic is similar to the logic in the comment action.

shine2lay: ah ok, i see it now.

shine2lay (approved):

lib/actions/close.js
samj1912: There were already some test cases for the single issue. I also added a test case for scheduler.

PR #389, opened 2020-09-29
fix: don't throw error if required status is not done

### LINKED ISSUE #386: merge action: add check for required status before attempting to merge
We are not checking required statuses currently before merging, as a result, when a new commit is pushed to PR and other required status (e.g circle ci) not finished, it'll throw error

shine2lay:
There weren't any API to check if all required status is done so the solution implemented here is to just not throw error if we see this specific type of error

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,7 @@
 CHANGELOG
 =====================================
 
+| October 2, 2020 : Don't throw merge error if required status are are the cause of the error `#389 <https://github.com/mergeability/mergeable/issues/389>`_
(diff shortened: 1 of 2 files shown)


samj1912:
This currently also fails in case one of the required branch checks is that the branch should be up to date or if the branch needs reviewers.

samj1912:
We should be checking for https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#get-the-combined-status-for-a-specific-reference, I can make a PR soon.

jusx (approved):

jusx:
@samj1912 Great catch. Looking forward to your PR.

shine2lay:
@jusx @samj1912 I'll merge this in for a temporary solution, left the issue open for when you open your PR.

PR #392, opened 2020-10-04
fix(docs): header in title validator doc

abid-mujtaba:
The doc is for the title validator but the top-level header says "Project".


jusx (approved):

jusx:
@abid-mujtaba Thank you for catching and fixing this.

PR #404, opened 2020-10-06
fix: Fix undefined error with blank validators

### LINKED ISSUE #402: feature request: add null validator
This arose from a discussion on #397. In certain use cases one wants to take an action on a specific event without any further validation (the event itself should trigger the event). For example, do something when a PR review is requested (triggered by `pull_request.review_requested`).

Currently all stanzas require a `validate` section. A solution is to introduce a **null** validator that does no actual validation, always passes, and so always triggers the relevant action.

### LINKED ISSUE #403: label validator: does not support `pull_request.review_requested` or `pull_request.closed`
Even though the [docs](https://mergeable.readthedocs.io/en/latest/validators/label.html) state that the Label validator supports `pull-request.*` the innards reveal that it does not (source: https://github.com/mergeability/mergeable/blob/master/lib/validators/label.js#L6).

This might be resulting in a `TypeError` I have been seeing:

> ❌ Error Occurred while executing an Action
If you believe this is an unexpected error, please report it on our issue tracker: https://github.com/mergeability/mergeable/issues/new
>
> Error Details
TypeError: Cannot read property 'schedule' of undefined

Is there a reason why the label validator should not be used with `pull_request.closed` or `pull_request.request_review`?

samj1912:
Fixes #402 
Fixes #403

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,7 @@
 CHANGELOG
 =====================================
 
+| October 6, 2020 : fix: Fix undefined error with blank validators `#402 <https://github.com/mergeability/mergeable/issues/402>`_
 | October 4, 2020 : fix Typo in header of title validator docs
 | October 2, 2020 : Don't throw merge error if required status are are the cause of the error `#389 <https://github.com/mergeability/mergeable/issues/389>`_
 | September 24, 2020 : Add ability to delete or replace the labels on an issue `#380 <https://github.com/mergeability/mergeable/issues/380>`_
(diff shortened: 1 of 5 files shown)


samj1912:
@jusx @shine2lay quick review please - mostly just docs :)

jusx (approved):
LGTM. I like the new example recipe you added.

jusx:
@samj1912 Would you mind adding tests for scenarios with the empty validationSuite array? I'll merge it if you push the added tests to this PR.

Let me know if you want to open a separate PR for the tests. I'd be OK with that as well.

PR #405, opened 2020-10-06
fix: Size validator - do not ignore hidden files by default

### LINKED ISSUE #401: Size validator: ignores hidden files by default
The Size validator uses `minimatch` with its default config ([link](https://github.com/mergeability/mergeable/blob/2338bde456a76f8d41e90b8706fcf84d7f7d25dc/lib/validators/size.js#L271)) which ignores hidden files by default. Since that includes `.github/mergeable.yml`, this led me on a merry chase for several hours.

I believe that the Size validator should validate against **all** files in PR by default, including hidden files. The user can use `match` and `ignore` to modify this default behavior. This can be resolved easily by using the `dot` configuration in `minimatch`: [doc](https://github.com/isaacs/minimatch#dot).

samj1912:
Fixes #401

### DIFF
--- a/__tests__/unit/validators/size.test.js
+++ b/__tests__/unit/validators/size.test.js
@@ -203,6 +203,32 @@ describe('PR size validator', () => {
     expect(validation.validations[2].status).toBe('pass')
   })
 
+  test('match hidden files', async () => {
+    const size = new Size()
+    const settings = {
+      do: 'size',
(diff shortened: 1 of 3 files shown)


samj1912:
@jusx again a really small fix complete with tests

jusx (approved):
Thanks! Looks good.

PR #409, opened 2020-10-08
test: Add tests for empty validators

samj1912:


samj1912:
@jusx as requested - tests for an empty validator

__tests__/unit/flex/flex.test.js
```diff
@@ -94,6 +94,64 @@ describe('Test processBeforeValidate and processAfterValidate invocations', () =
...
+        validate: []
+        pass:
+          - do: comment
+            payload:
+              body: >
+                Thanks for creating a pull request! A maintainer will review your changes shortly. Please don't be discouraged if it takes a while.`
+
+    Helper.mockConfigWithContext(context, config)
+    let registry = { validators: new Map(), actions: new Map() }
+    const commentAction = new Action()
```
jusx, line 112: The change in question where the logic to test the empty array are in each individual action that inherits `Action` -- for that reason, I don't feel this actually is testing the conditions. 

Would be good if the tests are validating the line in question:
`results && results.validationSuites && results.validationSuites.length && results.validationSuites[0].schedule`

Thoughts?

samj1912: @jusx fixed :)

jusx (approved):
Ideally the common logic in question is refactored out to a function or it's super class for better unit testing and coverage but I'm OK with this for now.

PR #417, opened 2020-10-10
chore: DRY schedule supported actions

jusx:
### Changes
- Abstracted common code for Actions that support the `schedule` event.
- Added tests for the above.

### DIFF
--- a/lib/actions/action.js
+++ b/lib/actions/action.js
@@ -47,6 +47,18 @@ class Action extends EventAware {
     }
     this.log.info(JSON.stringify(usageLog))
   }
+
+  /**
+   * Returns the actionable issues and/or pulls if it's for a schedule event. Otherwise return the current issue or pull that is for the current event.
+   */
+  getActionables (context, results) {
+    let scheduleResults = results && results.validationSuites && results.validationSuites.length && results.validationSuites[0].schedule
+    let actionables = (scheduleResults)
+      ? scheduleResults.issues.concat(scheduleResults.pulls)
+      : [this.getPayload(context)]
+
+    return actionables
+  }
 }
 
 module.exports = {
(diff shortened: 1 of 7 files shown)


lib/actions/action.js
```diff
@@ -47,6 +47,18 @@ class Action extends EventAware {
     }
     this.log.info(JSON.stringify(usageLog))
   }
+
+  /**
+   * Returns the actionable issues and/or pulls if it's for a schedule event. Otherwise return the current issue or pull that is for the current event.
+   */
+  getActionables (context, results) {
+    let scheduleResults = results && results.validationSuites && results.validationSuites.length && results.validationSuites[0].schedule
```
samj1912, line 55: We could instead check for `context.event === 'schedule'` and then assign this, as it sometimes fails for other events.

shine2lay: since this piece of code is verbatim refactoring, does the issue you mentioned currently exists in our production?

shine2lay (approved):
LGTM! you should follow up with @samj1912 comment before merging

lib/actions/action.js
samj1912: yup, although this works as expected, it's simply more explicit about what is expected to have these attributes. That way we can debug errors more easily. Currently in production it only fails if someone explicitly adds an event listener which we don't support for a particular action.

jusx: I prefer to be explicit as well. The suggestion to use `context.event` is a good one.

shine2lay (approved):

lib/actions/action.js
```diff
@@ -47,6 +47,20 @@ class Action extends EventAware {
...
     this.log.info(JSON.stringify(usageLog))
   }
+
+  /**
+   * Returns the actionable issues and/or pulls if it's for a schedule event. Otherwise return the current issue or pull that is for the current event.
+   */
+  getActionables (context, results) {
+    let scheduleResults = (context.event === 'schedule')
+      ? results && results.validationSuites && results.validationSuites.length && results.validationSuites[0].schedule
+      : null
```
samj1912, line 57: Should it return an empty array instead of null here so that actionables.map doesn't fail in case nothing is present?

samj1912 (approved):

PR #394, opened 2020-10-05
[wip] feat: Add the ability to auto merge PRs

samj1912:
This PR adds a couple of things - 

1. A new event type listener for `status.*` this allows us to listen to status change events so that we can fire the `merge` action on `status` change events.
2. A validator for `status.*`. This validator resolves the commit related to the status to relevant PRs on the repository. There is a `checks` setting option it takes to filter our PRs with failing checks. (This is optional as some checks may not be required to do merge)
3. Adds support for status* events to the merge actions.
4. Checks that the pull request is not in a `blocked` `mergeable_state` before trying to merge. This ensures that the PR is ready to merge (passes required checks including status, review and other branch protection checks)
5. Fixes broken `merge` action. `merge` action was missing the `name` parameter.

Fixes #395 

Fixes #386 

TODO: Update docs and add tests.

PR #418, opened 2020-10-11
feat: Add the ability to auto-merge pull requests

### LINKED ISSUE #395: feature request: add validator for other Github checks in a PR
My PRs often have long-running CI and linter checks. These don't work with the *merge* action in this bot since the bot has no way to validate and block based on other Github checks that are pending or have failed. This causes the bot to attempt to merge the PR and fail.

It would be really useful to have a validator that ensures that *other* Github checks have all passed which we can then use to automatically merge.

samj1912:
Fixes #395 

Better version of #394 

We need to add support for both `status.*` events and `check_suite.*` events since the former is used by CI like Jenkins/Circle CI and the latter is used by apps like `mergeable` and Github Actions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2 # use CircleCI 2.0
 jobs: # a collection of steps
   build: # runs not using Workflows must have a `build` job as entry point
     docker: # run the steps with Docker
-      - image: circleci/node:8.8.1 # ...with this image as the primary container; this is where all `steps` will run
+      - image: circleci/node:12.16.0 # ...with this image as the primary container; this is where all `steps` will run
     steps: # a collection of executable commands
       - checkout # special step to check out source code to working directory
       # - run:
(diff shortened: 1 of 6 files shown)


samj1912:
@jusx @shine2lay review please. I have been using this patch for a while and it works perfectly.

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2 # use CircleCI 2.0
 jobs: # a collection of steps
   build: # runs not using Workflows must have a `build` job as entry point
     docker: # run the steps with Docker
-      - image: circleci/node:8.8.1 # ...with this image as the primary container; this is where all `steps` will run
+      - image: circleci/node:12.16.0 # ...with this image as the primary container; this is where all `steps` will run
```
jusx, line 5: May I ask the reason for changing the version here?

jusx (commented):
This looks good. Interesting approach with the empty validator. Our original thought was to use a scheduler but I like this better. Nicely done.

I'll approve this as soon as I understand the reasoning behind the node version bump on circleci.

.circleci/config.yml
samj1912: Mostly because node 8 is no longer supported https://blog.risingstack.com/update-nodejs-8-end-of-life-no-support/

I believe that this change is already in master though?

jusx: You're right. :-)

jusx (approved):

PR #420, opened 2020-10-11
feat: Add a config cache

### LINKED ISSUE #407: Mergeable reads the mergeable config on each event
It needs to fetch the config for each event to determine if the event is a no op or not based on the config. This can be potentially resource consuming. We should instead have a cache and invalidate it if the config was changed.

samj1912:
Fixes #407

### DIFF
--- a/lib/configuration/configuration.js
+++ b/lib/configuration/configuration.js
@@ -88,7 +88,7 @@ class Configuration {
     }
   }
 
-  static async updateCache (context, cache) {
+  static async resetCache (context, cache) {
     // Only update the config cache in case there was a push
     // event that updated the config file
     if (context.event !== 'push') return
@@ -152,7 +152,7 @@ class Configuration {
     const configPath = process.env.CONFIG_PATH ? process.env.CONFIG_PATH : 'mergeable.yml'
     if (process.env.USE_CONFIG_CACHE) {
       const cache = Configuration.getCache()
-      Configuration.updateCache(context, cache)
+      Configuration.resetCache(context, cache)
       config = cache.get(`${repo.owner}/${repo.repo}`)
       if (_.isUndefined(config)) {
         config = await context.probotContext.config(configPath)
--- a/lib/configuration/configuration.js
+++ b/lib/configuration/configuration.js
@@ -1,5 +1,8 @@
 const yaml = require('js-yaml')
 const _ = require('lodash')
+const NodeCache = require('node-cache')
+
+const configCache = new NodeCache({ stdTTL: 0, maxKeys: 5000 })
 
 class Configuration {
   constructor (settings) {
@@ -10,7 +13,6 @@ class Configuration {
     }
 
     this.settings = settings
-
     this.validate()
     if (this.errors.size > 0) return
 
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -45,6 +46,7 @@
     ]
   },
   "jest": {
+    "testTimeout": 15000,
```
jusx, line 49: @samj1912 I assume you are timing out on localhost when running the tests?

lib/configuration/configuration.js
```diff
@@ -119,9 +141,18 @@ class Configuration {
...
     // probotContext.config loads config from current repo or from a repo called
     // '.github' in the same organisation as a fallback. It returns the parsed YAML object.
-
+    let config = null
     const configPath = process.env.CONFIG_PATH ? process.env.CONFIG_PATH : 'mergeable.yml'
-    const config = await context.probotContext.config(configPath)
+    if (configCache !== null) {
+      Configuration.updateCache(context)
+      config = configCache.get(`${repo.owner}/${repo.repo}`)
+      if (config === undefined) {
```
jusx, line 149: Since we have loadash in the codebase and you're using it on L97, I propose we stay consistent here and use `_.isUndefined`

lib/configuration/configuration.js
```diff
@@ -1,5 +1,8 @@
 const yaml = require('js-yaml')
 const _ = require('lodash')
+const NodeCache = require('node-cache')
```
jusx, line 3: I don't have a strong opinion here on choice of caching modules, but am curious on yours here.

lib/configuration/configuration.js
```diff
@@ -119,9 +141,18 @@ class Configuration {
...
+      Configuration.updateCache(context)
+      config = configCache.get(`${repo.owner}/${repo.repo}`)
+      if (config === undefined) {
+        config = await context.probotContext.config(configPath)
+        configCache.set(`${repo.owner}/${repo.repo}`, config)
+      }
+    } else {
+      config = await context.probotContext.config(configPath)
+    }
     if (config === null || config === undefined) {
```
jusx, line 156: I know this is old code you didn't introduce -- probably been around since before `loadash` was introduced into the codebase, but I think it's time we update this to use `_.isNull` and `_.isUndefined` for consistency.

jusx (commented):
Looks good. Just some minor nits.

jusx:
Please consider adding tests for the new caching code :-)

lib/configuration/configuration.js
samj1912: This was the most popular and well supported/recently updated cache module that support TTL.

package.json
samj1912: Removed.

samj1912:
@jusx this should be ready for review

lib/configuration/configuration.js
```diff
@@ -86,10 +88,30 @@ class Configuration {
     }
   }
 
+  static async updateCache (context, cache) {
```
jusx, line 91: I find the name of this function to be confusing. It is reseting/clearing the cache here and not necessarily updating the cache. The set occurs in `fetchConfigFile`. I believe it will be clearer to name this function something like `resetCache` 

Thoughts?

jusx (commented):
Looks good. Thanks for including the tests here! I left a minor comment. Let me know your thoughts.

lib/configuration/configuration.js
samj1912: Done

jusx (approved):

Issue #343, opened 2020-07-23
enforce stricter rules on certain branches


epDHowwD:
Would it be possible to use the app to enforce stricter rules for PRs targeting a limited set of base branches, e.g. `release` branch. Here I would like to enforce additional rules like title style for automated changelog generation, just as an example. Other branches, like `master` or `develop` might have less strict rules.

I am thinking of something like this:
``` yaml
  - when: pull_request.*, pull_request_review.*
    filter:
      - base: ['release']
```

shine2lay:
@epDHowwD I think this is a great idea :muscle: , it is currently not do-able in mergeable but I think this is a great candidate for the next feature to be added

moltob:
I was expecting the plugin to read it's configuration always from the base branch of a PR, hence allowing separate configurations for different branches this way, much like Github's codeownership is evaluated. But then I read [this comment](https://github.com/probot/probot/blob/c8f7f7374d85a600f1465d8bc6d0344d3bc63f75/src/context.ts#L219-L220) in the probot API...

I'd like to vote for this feature to be considered for implementation. The plugin is great and allows a very flexible definition of mergeability. Having that on branches work differently would be terrific.

jusx:
@epDHowwD Closing this since #421 is merged.

PR #421, opened 2020-10-12
feat: added BaseRef-validator based on Title-validator

roydenwa:
BaseRef-validator for pull_requests to check "destination branch" with must_include and must_exclude conditions

### DIFF
--- a/lib/validators/baseRef.js
+++ b/lib/validators/baseRef.js
@@ -1,0 +1,63 @@
+const { Validator } = require('./validator')
+
+class BaseRef extends Validator {
+  constructor () {
+    super('baseRef')
+    this.supportedEvents = [
+      'pull_request.opened',
+      'pull_request.edited',
+      'pull_request_review.submitted',
+      'pull_request_review.edited',
+      'pull_request_review.dismissed',
+      'pull_request.labeled',
+      'pull_request.milestoned',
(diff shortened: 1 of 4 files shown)


epDHowwD:
fixes: #343

lib/validators/baseRef.js
```diff
@@ -0,0 +1,43 @@
...
+      'pull_request_review.dismissed',
+      'pull_request.labeled',
+      'pull_request.milestoned',
+      'pull_request.demilestoned',
+      'pull_request.assigned',
+      'pull_request.unassigned',
+      'pull_request.unlabeled',
+      'pull_request.synchronize',
+      'pull_request.push_synchronize'
+    ]
```
jusx, line 20: When all the actions of an event is supported we typically like to write this like so:
```js
this.supportedEvents = [
      'pull_request.*',
      'pull_request_review.*'
    ]
```

This way, it is understood that everything within pull_request and pull_request_review is supported. It is becomes consistent with the docs.

jusx (commented):
LGTM -- will approve and merge after addressing the comment.

lib/validators/baseRef.js
roydenwa: @jusx makes sense, I changed it.

jusx (approved):

PR #425, opened 2020-10-22
docs: Identation in yml correction for greeting a new contributor.

jusx:
## Changes
- Minor indentation correction in recipe yml example for Greet a contributor.

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -142,7 +142,7 @@ Add a comment on a pull request when it is created
           - do: comment
             payload:
                 body: >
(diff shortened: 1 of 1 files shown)


docs/recipes.rst
```diff
@@ -142,7 +142,7 @@ Add a comment on a pull request when it is created
           - do: comment
             payload:
                 body: >
```
samj1912, line 144: Shouldn't this be dedented instead? This seems to have an indentation of 4, when the rest of the file is 2. (Also probably my mistake, sorry)

jusx: Good catch. I'll update.

samj1912 (approved):

shine2lay (approved):

PR #434, opened 2020-12-11
feat: Helm Chart implemantation for kubernetes

iammuho:
Fixes : https://github.com/mergeability/mergeable/issues/435

**Problem:**

There is no easy way to install Mergeable into the Kubernetes cluster.

**Solution:**

This PR implements a helm-chart to install mergeable into the Kubernetes cluster.

### DIFF
--- a/helm/mergeable/values.yaml
+++ b/helm/mergeable/values.yaml
@@ -10,7 +10,7 @@ namespace:
   labels: {}
 
 image:
-  repository: 489198589229.dkr.ecr.eu-west-1.amazonaws.com/mergeable
+  repository: "DOCKER_REPO/mergeable"
   pullPolicy: IfNotPresent
   tag: ""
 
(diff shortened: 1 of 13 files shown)


shine2lay:
@jusx can you please review this? I am out of my depth here.

helm/mergeable/values.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+replicaCount: 1
+
+namespace:
+  enabled: true
+  annotations: {}
+  labels: {}
+
+image:
+  repository: "DOCKER_REPO/mergeable"
```
jusx, line 13: Is DOCKER_REPO suppose to be an env variable?

iammuho: it's supposed to be filled, no need to get from environment values I guess, but just static typing

jusx (approved):
LGTM but I have not tested it and I assume you have.

Issue #429, opened 2020-11-16
Is it possible to re-order the default order for configuration files?


iammuho:
As I can understand from the documents;

```
Mergeable will now use this file as the default when it cannot find one in a given repository or PR. It determines the file to use in the following order:
A mergeable.yml inside the PR if the PR originates from the same repository. If it originates from a fork it is not loaded for security reasons. See #406 for more details.
A mergeable.yml inside the repository the PR is for.
A mergeable.yml at <YOUR_ORG>/.github/.github/mergeable.yml.
```

But there is a security risk for this text, if I've configured the `mergeable` in global or master branch by saying just this labels can approve the PR, any other PR owner can override this rule. Am I right?

If yes, is there any option to change this ? I want order to be `global > base branch> pr`

shine2lay:
@iammuho I am not sure I fully understand your thought process. 
First, To answer your second question 

`is there any option to change this ? `
the answer is no, we don't have a config to change the default

Back to the first part,
By PR owner, can I assume you mean someone using a different fork to create a PR on your repo?
If this is the case, then the new mergeable config from the PR wouldn't load because of security reason. see #406 

If PR originate from the same repo, then the new config would be loaded when running mergeable, this is helpful in testing if mergeable works before merging the PR. So technically, anyone that has a write permission to the repo have the ability to change the `mergeable` config. 

does this answer your first question? If not, can you elaborate a bit further?

jusx:
I believe you can have settings where only certain configured users or admins can override and merge a PR that is not green.

@iammuho Are you concerned that allowing a PR to override global configurations can be problematic since the author  can circumvent certain processes that is in place.

Would like to understand your use case. Could you elaborate?

@shine2lay While #468 works for self installed to prevent this, we may want to consider a configuration option in the yml. Thoughts?

iammuho:
Hey @jusx and @shine2lay,

As you can see my friends PR ,https://github.com/mergeability/mergeable/pull/436, this configuration is for not fork project PR but internal repo PR flow. 

Let's think about our organization,

in HelloFresh, we have created an organization standard and wish PRs to be checked by this config. And lets assume the description shouldn't be empty. But developers can easily bypass that by just sending a `mergeable.yml` file in the PR and PR will be checked in this new `mergeable.yml` which is not suitable.

[...]

PR #436, opened 2020-12-11
feat: Add env to control which configuration file to use

diegomarangoni:
Fixes: #429

**Problem:**
Currently there is no way to control which mergeable configuration file will be used, if the one from the PR branch or the one from the default branch.

**Solution:**
This PR implements a new environment variable to control that while maintaining the previous behaviour

### DIFF
--- a/lib/configuration/configuration.js
+++ b/lib/configuration/configuration.js
@@ -112,7 +112,13 @@ class Configuration {
   static async fetchConfigFile (context) {
     let github = context.github
     let repo = context.repo()
-    if (['pull_request', 'pull_request_review'].includes(context.event)) {
+    // using pull request configuration by default, this behaviour
+    // can be controlled using USE_CONFIG_FROM_PULL_REQUEST environment variable
+    let usePullRequestConfig = true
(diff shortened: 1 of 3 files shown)


lib/configuration/configuration.js
```diff
@@ -112,7 +112,13 @@ class Configuration {
   static async fetchConfigFile (context) {
     let github = context.github
     let repo = context.repo()
-    if (['pull_request', 'pull_request_review'].includes(context.event)) {
+    // using pull request configuration by default, this behaviour
+    // can be controlled using USE_CONFIG_FROM_PULL_REQUEST environment variable
+    let usePullRequestConfig = true
+    if (process.env.USE_CONFIG_FROM_PULL_REQUEST !== undefined && process.env.USE_CONFIG_FROM_PULL_REQUEST === 'false') {
```
shine2lay, line 118: could you please add a test case for this?

shine2lay (changes requested):

shine2lay (approved):

shine2lay:
@diegomarangoni could you please resolve the conflict? I am not able to do it.

diegomarangoni:
@shine2lay conflict resolved

Issue #438, opened 2020-12-14
There is no way to validate branch name with regex


iammuho:
Problem:

As a repo owner, I want to validate branches of the Pull-requests to have a format that is defined by regex.

Solution:

We need to have a `headRef` validator just as same as `baseRef`

jusx:
@shine2lay This is potentially a duplicate of #424 -- fyi

shine2lay:
duplicate of #424

PR #439, opened 2020-12-14
feat: headref validator

iammuho:
Issue: https://github.com/mergeability/mergeable/issues/438

---

With this PR, developers can validate headRef (branch) with regex tags.

Sample:

```
- do: headRef
        must_include:
           regex: '(^|[-\/])(major|release|minor|feature|patch|issue|hotfix|dependabot\/|whitesource)[-\/](\b[A-Z][A-Z0-9]+-\d+\b).+$'
           message: 'Your branch name should be like feature/SOME-TICKET-NUMBER-some-description`
```

### DIFF
--- a/lib/validators/headRef.js
+++ b/lib/validators/headRef.js
@@ -1,7 +1,7 @@
 const { Validator } = require('./validator')
 
 class HeadRef extends Validator {
-  constructor() {
+  constructor () {
     super('headRef')
     this.supportedEvents = [
       'pull_request.*',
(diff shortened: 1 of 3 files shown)


lib/validators/headRef.js
```diff
@@ -0,0 +1,32 @@
+const { Validator } = require('./validator')
+
+class HeadRef extends Validator {
```
shine2lay, line 3: could you please add tests for this validator?

shine2lay (changes requested):

shine2lay (approved):
this is missing documentation, but i'll go ahead and fix that in a separate PR

Issue #442, opened 2020-12-17
Team member listing for organization has only one page results


iammuho:
**Problem:**

If you `limit` approval validation based on the organization team and not CODEOWNERS file, Mergeable is trying to `listMembersInOrg` without a pagination and that results to list only 30 users. And this is a problem for companies who has more than that.

**Solution:**

- We can simply use `paginate` module of the OCTOKIT to fetch every page but that will result so much performance issues if user has defined multiple teams and teams total has a few pages.
- We can change alghoritm to rather than Listing all members, keeping in array and search in array, we can use `getMembershipInOrg` to fetch the approval user membership status in teams, if one of them returns 200 - active, we can return true or false.

jusx:
Resolved by #443

PR #443, opened 2020-12-17
fix: member listing pagination issue

iammuho:
Related Issue: https://github.com/mergeability/mergeable/issues/442

-- 

This PR follows the second idea which I've changed the alghorithm to fetch approved users membership states and if any of them are true, it will be counted.

### DIFF
--- a/lib/validators/options_processor/teams.js
+++ b/lib/validators/options_processor/teams.js
@@ -91,11 +91,7 @@ const getTeamMembershipState = async (context, team, username) => {
     throw err
   }
 
-  if (res.data.state === 'active') {
-    return true
-  }
-
-  return false
+  return (res.data.state === 'active')
 }
 
 module.exports = Teams
--- a/lib/validators/options_processor/teams.js
+++ b/lib/validators/options_processor/teams.js
@@ -20,9 +20,8 @@ class Teams {
     return _.uniq(teamMembers)
   }
 
-  // extractTeamMemberships returns a list of active state users from given teams/users 
-  static async extractTeamMemberships(context, teams, users) {
-    let teamMembers = []
+  static async extractTeamMemberships (context, teams, users) {
+    const teamMembers = []
     let membershipState = false
 
     if (!teams || teams.length === 0) return teamMembers
@@ -87,7 +86,7 @@ const getTeamMembershipState = async (context, team, username) => {
     })
   } catch (err) {
     if (err.status === 404) {
-       return false
+      return false
     }
     throw err
   }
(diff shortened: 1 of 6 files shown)


lib/validators/options_processor/teams.js
```diff
@@ -19,6 +19,26 @@ class Teams {
...
+        }
+
+        if (membershipState) { teamMembers.push(user) }
+      }
+    }
+    return _.uniq(teamMembers)
+  }
 }
 
 const getTeamMembers = async (context, team) => {
```
shine2lay, line 44: i see that `[owner](https://github.com/mergeability/mergeable/blob/master/lib/validators/options_processor/owners.js#L37)` file still uses this function, is there a particular reason for it?

shine2lay (changes requested):

lib/validators/options_processor/teams.js
```diff
@@ -47,4 +67,35 @@ const getTeamMembers = async (context, team) => {
...
+      username: username
+    })
+  } catch (err) {
+    if (err.status === 404) {
+      return false
+    }
+    throw err
+  }
+
+  if (res.data.state === 'active') {
```
jusx, line 94: You can replace L94 - 98 in a succinct manner:

` return (res.data.state === 'active')`

lib/validators/options_processor/teams.js
```diff
@@ -47,4 +67,35 @@ const getTeamMembers = async (context, team) => {
   return res.data.map(member => member.login)
 }
 
+const getTeamMembershipState = async (context, team, username) => {
+  const stringArray = team.split('/')
+  if (stringArray.length !== 2) {
+    throw Error(`each team id needs to be given in 'org/team_slug'`)
+  }
+  if (stringArray[0].indexOf('@') === 0) stringArray[0] = stringArray[0].substring(1)
```
jusx, line 75: Looks like you are supporting both naming prefix with `@` and without here. i.e.`@org/teamName` and `org/teamName` here. 

Good call.

To remove the first element, you may want to consider using `stringArray.shift()` here.

jusx:
Love the performant solution to this. Thank you.

lib/validators/options_processor/teams.js
iammuho: In fact, I haven't add this but just copy/paste from old code. Do you want me to replace again ?

jusx: I do think it's more readable with `shift`. If you agree, please do. Not a big deal otherwise.

iammuho:
hey @jusx ,

I've used the shift, can you re-review?

best

jusx (approved):

jusx:
@iammuho I went ahead and merged this. It would be great if you are able to do a separate PR to add code coverage on the new function in `Teams` Thanks.

PR #445, opened 2020-12-17
feat: Add docker image build and push

### LINKED ISSUE #427: Setup docker image
Would be nice to have a `Dockerfile` on the repo, and also to setup an Actions pipeline to build and push to Docker Hub.

The Dockerfile below is a working example.
```Dockerfile
FROM node:14

RUN npm install -g probot

WORKDIR /app

COPY package*.json ./

RUN npm ci --only=production

COPY . .

EXPOSE 3000

USER 1000:1000

CMD [ "probot", "run", "./index.js" ]
```

diegomarangoni:
Fixes #427

This PR introduces a Docker image by adding a Dockerfile and an Actions pipeline to build and push to Docker Hub.

A Docker Hub account will be required and also is necessary to set repository secrets `DOCKERHUB_USERNAME` with the account name and `DOCKERHUB_PASSWORD` with the generated token.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,11 +7,7 @@ RUN npm ci --only=production
 
 COPY . .
 
-FROM gcr.io/distroless/nodejs:14
-
-COPY --from=build /app /app
-
-EXPOSE 3000
+ENV PORT=${PORT:-3000}
 USER 1000:1000
 
-CMD [ "/app/node_modules/probot/bin/probot.js", "run", "/app/index.js" ]
+CMD ./node_modules/probot/bin/probot.js run --port $PORT ./index.js
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14 AS build
+FROM node:12.16.0
 
 WORKDIR /app
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM node:14 AS build
+
+WORKDIR /app
+
+COPY package*.json ./
+RUN npm ci --only=production
+
+COPY . .
+
+FROM gcr.io/distroless/nodejs:14
+
+COPY --from=build /app /app
+
+EXPOSE 3000
+USER 1000:1000
+
+CMD [ "/app/node_modules/probot/bin/probot.js", "run", "/app/index.js" ]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+npm-debug.log
+*.pem
+.env
+coverage
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,44 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - '*'
+    branches:
+      - master
+
+jobs:
+  build:
+    name: Build and push Docker image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+COPY package*.json ./
+RUN npm ci --only=production
+
+COPY . .
+
+FROM gcr.io/distroless/nodejs:14
+
+COPY --from=build /app /app
+
+EXPOSE 3000
```
jusx, line 14: Do you envision this dockerfile to be used for development purposes on locahost or a generic one for deployment to containers?

If it's the second, should the port come from an env var? And [in the case of Heroku](https://help.heroku.com/PPBPA231/how-do-i-use-the-port-environment-variable-in-container-based-apps), you'd have to cater to its' nuances.

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM node:14 AS build
+
+WORKDIR /app
+
+COPY package*.json ./
+RUN npm ci --only=production
+
+COPY . .
+
+FROM gcr.io/distroless/nodejs:14
```
jusx, line 10: I noticed that you are specifying node 14+ here and I'm assuming it's because it's the latest stable version. 

I would prefer that the version align with that specified in the [CI](https://github.com/mergeability/mergeable/blob/master/.circleci/config.yml) since that's what we run our test suite against. Thoughts?

jusx (commented):
Thanks for the PR. Left a couple of minor comments.

diegomarangoni:
> Thanks for the PR. Left a couple of minor comments.

I pinned the nodejs image version and also add the `$PORT`env variable

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+COPY package*.json ./
+RUN npm ci --only=production
+
+COPY . .
+
+ENV PORT=${PORT:-3000}
+USER 1000:1000
+
+CMD ./node_modules/probot/bin/probot.js run --port $PORT ./index.js
```
aleksandrzhiliaev, line 13: Would it be better to just leave `ENTRYPOINT` and do `RUN` to ensure that Dockerfile is valid?
Something like:
```
ENTRYPOINT ["./node_modules/probot/bin/probot.js"]

# ensure that Dockerfile is valid
RUN ["./node_modules/probot/bin/probot.js", "--version"]
```
Not sure that `probot.js` has --version command, it is just an idea.

diegomarangoni: I think it might not work on Heroku using `$PORT` environment variable.
To variable expansion works properly I need to run using `shell` form, and not `exec`.

jusx: @diegomarangoni Looks good and I've verified everything works. I recommend adding a `dockerignore` to avoid accidental inclusion of secrets in the image by adding `.env` and `*.pem` to it. Once you add that and update your branch, we are good to merge.

Additionally if you want to, you can `dockerignore` files not needed like docs and images to ensure optimal image size. Though I'm totally fine with image size optimization being a separate PR.

jusx:
@diegomarangoni Looks good and I've verified everything works. I recommend adding a `dockerignore` to avoid accidental inclusion of secrets in the image by adding `.env` and `*.pem` to it. Once you add that and update your branch, we are good to merge.

Additionally if you want to, you can `dockerignore` files not needed like docs and images to ensure optimal image size. Though I'm totally fine with image size optimization being a separate PR.

diegomarangoni:
@jusx nice! I did the dockerignore similar to gitignore (except for node_modules)

jusx (approved):

PR #460, opened 2021-01-11
chore: push to ghcr instead

LeoQuote:
In #445 a workflow was set up to push image to docker.io, while nobody has set up a corresponding dockerhub token or password for it.
I think we can just push it to ghcr.io instead, which does not require any configuration, it just works!

Update: I tried myself and found this workflow also requires a github personal token that has permission to push. So it's not exactly no configuration.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,44 +1,77 @@
-name: Release
+name: Docker
 
 on:
   push:
-    tags:
-      - '*'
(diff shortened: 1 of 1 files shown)


shine2lay:
@jusx i know you were planning to create a dockerhub account, what do you this approach?

LeoQuote:
@shine2lay I'm going to close this pr cause I found that this approach is not completely **no-configuration**, it also requires a GitHub personal token with sufficient permission.

PR #462, opened 2021-01-12
docs: add `and`, `or` validator to configuration tree

shine2lay:
Context
-----
`And` and `Or` validator are not currently showing up in readTheDocs because it is not properly added to configuration tree


docs/configuration.rst
```diff
@@ -75,6 +76,7 @@ Validator List
     validators/headRef.rst
     validators/label.rst
     validators/milestone.rst
+    validators/or.rst
```
jusx, line 79: Thoughts on adding these under advanced operators?

shine2lay: you mean a page called `advanced operators?(validator?)` and put these in them?

jusx: Yes. These are operators. So on the side bar you'd have `Validators`, `Operators` and `Actions`

shine2lay: addressed

jusx: Since you've added the Operators section and listed the operators there, should it still be on the validator list?

shine2lay: you are right, it was an oversight, fixed

jusx (approved):

PR #471, opened 2021-01-15
feat: Organization-wide as default config

### LINKED ISSUE #470: Support use organization-wide configuration as default
Right now, Mergeable is configured to use configuration file from repo itself or as a fallback, organization. But that causes problems to override our organization standarts by placing an empty `mergeable.yml` files into any repo.

To fix this problem,

We should define `org` + `local` mergeable configurations together.

`org` rules beat `local` rules always

Example:

**ORG**
- Place a Jira ticket ID into branch (i.e ees-4646-test)
- Description cannot be empty
- One developer should approve

**Local:**
- You should define a label for each PR

**Current:**
- You should define a label for each PR

**Should Be:**
- Place a Jira ticket ID into branch (i.e ees-4646-test)
- Description cannot be empty
- One developer should approve
- You should define a label for each PR

diegomarangoni:
fixes #470

This PR adds a new environment variable `USE_ORG_AS_DEFAULT_CONFIG` to control the behavior of inheriting organization-wide conf as default.


shine2lay:
@diegomarangoni I really like the idea, I think it's very useful. what do you think about this feature being defined within the configuration itself instead of an environment variable to be more flexible?

the config could be something like this 

```
version: 2
mergeable:
  - when : ...
config_settings:
  merge_org_and_local: true
```

lib/configuration/configuration.js
```diff
@@ -156,21 +156,46 @@ class Configuration {
...
       }
-    } else {
-      config = await context.probotContext.config(configPath)
     }
+
+    let orgConfig = null
+    let useOrgAsDefaultConfig = false
+    if (process.env.USE_ORG_AS_DEFAULT_CONFIG !== undefined && process.env.USE_ORG_AS_DEFAULT_CONFIG === 'true') {
+      useOrgAsDefaultConfig = true
+    }
```
jusx, line 173: You can write this succinctly with
```js
let useOrgAsDefaultConfig = (process.env.USE_ORG_AS_DEFAULT_CONFIG !== undefined && process.env.USE_ORG_AS_DEFAULT_CONFIG === 'true')
```

__tests__/unit/configuration/configuration.test.js
```diff
@@ -688,10 +689,53 @@ describe('with version 1', () => {
...
+            - do: title
+              must_include:
+                regex: 'nothing'`
+
+    let context = createMockGhConfig(repoConfig, orgConfig)
+    context.event = 'pull_request'
+
+    let config = null
+
+    process.env.USE_ORG_AS_DEFAULT_CONFIG = 'true'
```
jusx, line 729: There is an `.env.template` that is sorely outdated. Would be great if you can start keeping that updated by adding this env variable in with an explanation. We can resolve the other missing env variables separately.

jusx:
> @jusx any thoughts?

@shine2lay 
In general I think that there should be a way to configure mergeable settings. It can be something like:

1. If an organization wide `.github` repo with `mergeable.settings.yml` then it should use it.
2. if an organization wide settings do not exist, then it will use the one in path `<owner>/<repo>/.github`

I'd name the node `settings` so that it can be expanded:

```yml
settings:
   configuration:
      merge_organization_with_repo: true
      pull_request_override: false
   something_else:
   ...
```   
I think it's fine for it being in env for now but I agree that in the long term it should be multi-tenant.

diegomarangoni:
I like the idea @shine2lay @jusx
This can be done as a new feature and then all applicable ENVs moved to the configuration.

jusx:
> This can be done as a new feature and then all applicable ENVs moved to the configuration.

Agreed.

jusx (approved):

Issue #482, opened 2021-01-19
Jira Validation for Ticket Numbers


iammuho:
We need an option or validation to validate given ticket numbers in any validator like title or headRef via JIRA API Calls.

PR #481, opened 2021-01-19
chore: Async Options

iammuho:
Issue: https://github.com/mergeability/mergeable/issues/480

PR #483, opened 2021-01-19
feat: Jira Validator Option

iammuho:
Issue: https://github.com/mergeability/mergeable/issues/482
Note: For this option to be working, options should be running on async. Before this PR to be accepted, please look at this PR: https://github.com/mergeability/mergeable/pull/481 (MERGED)
---

This PR adds a `Jira` option to the below validators. If anyone wants to check JIRA TICKET to be validated through JIRA APIs;

1. Enable JIRA Settings via env variables;

```
JIRA_PASSWORD //API TOKEN
JIRA_USERNAME //USERNAME
JIRA_HOST //HOST
JIRA_PROTOCOL // default 'https'
JIRA_VERSION // default '2'
JIRA_STRICT_SSL // default 'true'
```

2. Enable JIRA VALIDATION option in validators;

```
---
version: 2
mergeable:
    name: 'checkfirst'
    validate:
      - do: title
        jira:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'JIRA Ticket ID is not valid'
```

You can use JIRA VALIDATION option in below validators;

- title
- headRef
- description
- label
- commit
- milestone


iammuho:
hey @jusx any comments for this ?

lib/validators/options_processor/options/jira.js
```diff
@@ -0,0 +1,82 @@
...
+
+const ENABLED_NOT_FOUND_ERROR = `Failed to run the test because 'enabled' is not provided for 'jira' option. Please check README for more information about configuration`
+
+class Jira {
+  static async process (validatorContext, input, rule) {
+    let isMergeable
+    let regexObj
+
+    const filter = rule.jira
+    const enabled = filter['enabled']
```
jusx, line 11: The enabled sub-option seem redundant to me. Maybe I'm missing something. Do you have a use-case where you have the `jira` option configured but you want to disable it? Wouldn't that be the same as just removing the `jira` option altogether?

jusx (commented):
Overall looks good. Left a comment about the enabled suboption.

iammuho:
hey @jusx , you can check again. You are right, no need for enabled.

lib/validators/commit.js
```diff
@@ -17,6 +17,12 @@ class Commit extends Validator {
       'pull_request_review.*'
     ]
     this.supportedSettings = {
+      jira: {
+        enabled: 'boolean',
```
jusx, line 21: We should remove all references to `enabled` since we've agreed it's not relevant.

__tests__/unit/validators/options_processor/options/jira.test.js
```diff
@@ -0,0 +1,88 @@
...
+  expect(checkTicketStatus).toHaveBeenCalledTimes(0)
+  expect(res.status).toBe('fail')
+})
+
+test('return fail if JIRA Ticket ID couldn\'t be found', async () => {
+  // Mock wrong JIRA Call
+  const checkTicketStatus = jest.fn().mockReturnValue(false)
+  jira.checkTicketStatus = checkTicketStatus
+
+  const rule = { jira: { enabled: true, regex: 'test' } }
```
jusx, line 71: Same as https://github.com/mergeability/mergeable/pull/483/files#r566509892 as well as in all the validators and the tests.

jusx (commented):
@iammuho Added additional comments. I would recommend including some docs either in this PR or create an issue regarding docs so that it is in the backlog.

iammuho:
@jusx , removed all references and sent the docs. I think in the docs we should also enable `options` menu to explain options one by one like validators and actions, because i.e JIRA option does work only self-hosted (to set environment values) For now, I've added how to add JIRA option to required validators, on another PR I'll add `options` part to the docs.

jusx (approved):

PR #485, opened 2021-01-21
chore: rename eventId to event_id for consistency

shine2lay:
rename eventId in logs to event_id for consistency

### DIFF
--- a/lib/actions/action.js
+++ b/lib/actions/action.js
@@ -31,7 +31,7 @@ class Action extends EventAware {
   logBeforeValidateUsage (context, settings) {
     const usageLog = {
       log_type: logger.logTypes.ACTION_BEFORE_VALIDATE_EXECUTE,
-      eventId: context.eventId,
+      event_id: context.eventId,
       repo: context.payload.repository.full_name,
       action_name: this.name,
       settings: JSON.stringify(settings)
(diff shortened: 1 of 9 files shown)


lib/actions/action.js
```diff
@@ -31,7 +31,7 @@ class Action extends EventAware {
   logBeforeValidateUsage (context, settings) {
     const usageLog = {
       log_type: logger.logTypes.ACTION_BEFORE_VALIDATE_EXECUTE,
-      eventId: context.eventId,
+      event_id: context.eventId,
```
jusx, line 34: We've been using camelCase throughout the codebase except for these properties which is snake case. If we want to be consistent, should all the properties in this JSON be changed to camelCase instead?

shine2lay: what are your thoughts? i am okay with either, i just notice this one is out of place. Our log types are in `snake_case` too, maybe we should change those too

jusx: LogTypes are constants in the traditional sense are all caps and because of that snake case. I'd leave those alone.

I prefer to be consistent with camelcase for the properties since that has been widely the convention.

shine2lay: sure, i'll change them

shine2lay:
@jusx done addressing comments, PTAL

jusx (approved):

PR #489, opened 2021-01-21
fix: bug where HttpError is returning for deepvalidation

### LINKED ISSUE #488: Bug: unknown_error_validator --- HttpError not found
image

this is occurring to multiple repos,
from a first glance, it could be `milestone` validator, will investigate more

Update: the error stems from deepvalidation component of the milestone validator

shine2lay:
fixes #488

### DIFF
--- a/lib/validators/options_processor/deepValidation.js
+++ b/lib/validators/options_processor/deepValidation.js
@@ -32,8 +32,7 @@ class DeepValidation {
             validator_name: property
           }
 
-          this.log.info(JSON.stringify(errorLog))
-          throw err
+          this.log.error(JSON.stringify(errorLog))
         }
         continue
       }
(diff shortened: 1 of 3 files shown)


lib/validators/options_processor/deepValidation.js
```diff
@@ -18,7 +19,24 @@ class DeepValidation {
...
+        // simply skip if the error is 404 (Not Found)
+        if (err.status !== 404) {
+          const errorLog = {
+            log_type: logger.logTypes.ISSUE_GET_FAIL_ERROR,
+            event_id: context.eventId,
+            repo: context.payload.repository.full_name,
+            validator_name: property
+          }
+
+          this.log.info(JSON.stringify(errorLog))
```
jusx, line 35: Should this be `log.error` since it's an unknown error that can potentially point to a bug.

lib/validators/options_processor/deepValidation.js
```diff
@@ -18,7 +19,24 @@ class DeepValidation {
...
+          throw err
```
jusx, line 36: I don't think we should throw here since you are already logging the error.

shine2lay: addressed

lib/validators/options_processor/deepValidation.js
shine2lay: addressed

jusx (approved):

Issue #486, opened 2021-01-21
Bug: HttpError: Review cannot be requested from pull request author.


shine2lay:
![image](https://user-images.githubusercontent.com/5243508/105282213-aea67800-5b62-11eb-8b2f-0e28b7798c16.png)

PR #490, opened 2021-01-21
fix: request review failing when reviewer is author of PR

shine2lay:
fixes: #486


lib/actions/request_review.js
```diff
@@ -1,10 +1,27 @@
...
 const _ = require('lodash')
 
 const createRequestReview = async (context, number, reviewers) => {
-  return context.octokit.pulls.requestReviewers(
-    context.repo({ pull_number: number, reviewers })
-  )
+  let res
+  try {
+    res = await context.octokit.pulls.requestReviewers(
+      context.repo({ pull_number: number, reviewers })
```
jusx, line 9: I feel we should filter out the author of the PR from the reviewers list.

lib/actions/request_review.js
```diff
@@ -1,10 +1,27 @@
...
+    // skip when error is "Review can not be requested from pull request author"
+    if (!err.message.toLowerCase().includes('pull request author')) {
+      const errorLog = {
+        log_type: logger.logTypes.REQUEST_REVIEW_FAIL_ERROR,
+        eventId: context.eventId,
+        repo: context.payload.repository.full_name,
+        action_name: this.name
+      }
+
+      this.log.info(JSON.stringify(errorLog))
```
jusx, line 21: If we filter out the author of the Pr from the reviewers list as per recommended on L9, then this should probably be a `log.error` since it's an unknown. and unexpected error.

jusx (commented):
Left some comments, let me know your thoughts, otherwise looks good as it does prevent failure.

shine2lay:
@jusx comments address, PTAL

lib/actions/request_review.js
```diff
@@ -1,10 +1,27 @@
...
-  )
+const createRequestReview = async (context, number, reviewers, actionObj) => {
+  let res
+  try {
+    res = await context.octokit.pulls.requestReviewers(
+      context.repo({ pull_number: number, reviewers })
+    )
+  } catch (err) {
+    // skip when error is "Review can not be requested from pull request author"
+    if (!err.message.toLowerCase().includes('pull request author')) {
```
jusx, line 13: This seem redundant, since the author is being filtered out.

shine2lay: good point, removed

jusx (approved):

PR #491, opened 2021-01-21
feat: If expression evaluation

diegomarangoni:
Currently, there is no way to apply a validator based on repository conditions (e.g if is public/private).
This PR adds a feature that's add a new field called `if` which contains a string expression that is evaluated.

```
version: 2
mergeable:
    name: 'repository rules'
    validate:
      - do: headRef
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Branch name contains Jira ticket (i.e ABC-123)'

      - do: title
        if: 'payload.repository.private == true && !("my-topic" in getRepositoryTopics())'
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Title contains Jira ticket (i.e ABC-123)'
```

Screenshot 2021-01-21 at 17 04 21


shine2lay:
@diegomarangoni I think we have a more flexible version of this if we have a new class that check for conditions on whether or not a recipe should be executed. What do you think about this type of config?

```
version: 2
mergeable:
    name: 'repository rules'
    filter:
      - check: repo
        isPrivate: true
        topic:
          regex: 'my topic'
    validate:
      - do: headRef
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Branch name contains Jira ticket (i.e ABC-123)'

      - do: title
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Title contains Jira ticket (i.e ABC-123)'
```

Where filters will be checked and only when true, the validations will run. I think filters can be extended to 3rd party integrations later as well.
@jusx

diegomarangoni:
I kinda like, feels more like the mergeable way.
what about this?
```yaml
version: 2
mergeable:
    name: 'repository rules'
    filter:
      - do: payload.repository
        visibility: private # can be private or public
      - do: repository.topics
        contains: 'my topic' # or can be an array [ 'my topic 1', 'my topic 2' ]
    validate:
      - do: headRef
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Branch name contains Jira ticket (i.e ABC-123)'
      - do: title
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Title contains Jira ticket (i.e ABC-123)'
```

shine2lay:
@diegomarangoni is there a reason you chose `payload.repository` over `repository`? 
IMHO they add a layer of unnecessary complexity, maybe you have other use cases you are thinking of?

diegomarangoni:
the idea here is to give the user the ability to do a condition based on any field of webhook `payload`
so users not only are able to check `repository` but also `organization`, `sender`, `pull_request`.

or I could do a filter for each of then, wdyt?

shine2lay:
If there is no other place for `repository` to be beside `payload`, I think using `repository` makes more sense because users don't need to care about payload itself.
and just to keep the consistency within mergeable, I think we should group `topics` inside `repository`. 

Something like this:
```
version: 2
mergeable:
    name: 'repository rules'
    filter:
      - do: repository
        visibility: private # can be private or public
        topics:
          regex: 'my topic' #using regex here instead of contains so it is more explicit that you can use regexExp to match
    validate:
      - do: headRef
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Branch name contains Jira ticket (i.e ABC-123)'
      - do: title
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Title contains Jira ticket (i.e ABC-123)'
```
wdyt?

diegomarangoni:
what if I want to check if the repository contains two or more topics at the same?
and what if I want to exclude repositories that contain one or more topics?

diegomarangoni:
what if something like this?

```yaml
version: 2
mergeable:
    name: 'repository rules'
    filter:
      - do: repository
        visibility: private # can be private or public
        topics:
          must_include: [ 'my topic' ]
          must_exclude: [ 'other topic' ]
    validate:
      - do: headRef
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Branch name contains Jira ticket (i.e ABC-123)'
      - do: title
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Title contains Jira ticket (i.e ABC-123)'
```

shine2lay:
i like it, I do have one question, for `must_include` is it "all topic must be present" or "at least one topic must be present"? it's a bit ambiguous to me


What about if follow the same pattern as the validators, then we can even re-use the code potential.
I mean something like this
```
version: 2
mergeable:
    name: 'repository rules'
    filter:
      - do: repository
        visibility: private # can be private or public
        topics:
          must_include: 
            regex : 'topic|second topic'
          must_exclude: 
            regex:  'this topic|other topic' 
    validate:
      - do: headRef
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Branch name contains Jira ticket (i.e ABC-123)'
      - do: title
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Title contains Jira ticket (i.e ABC-123)'
```

for the case where multiple topics must be included, the `and` option could come in handy

```
version: 2
mergeable:
    name: 'repository rules'
    filter:
      - do: repository
        visibility: private # can be private or public
        topics:
         and:
            - must_include: 
                 regex : 'first topic'
            - must_include: 
                regex:  'second topic' 
    validate:
      - do: headRef
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Branch name contains Jira ticket (i.e ABC-123)'
      - do: title
        must_include:
          regex: '[A-Z][A-Z0-9]+-\d+'
          regex_flag: none
          message: 'Title contains Jira ticket (i.e ABC-123)'
```

diegomarangoni:
IMHO regex does not work well when the field is an array.
it does work great when the field is a string, but not for an array

well, both ways will work for my use cases, so, whatever you prefer.

> i like it, I do have one question, for must_include is it "all topic must be present" or "at least one topic must be present"? it's a bit ambiguous to me

to solve this, we could use the terms `all` and `none`
```yaml
    filter:
      - do: repository
        visibility: private # can be private or public
        topics:
          all: [ 'my topic' ]
          none: [ 'other topic' ]
```

I working into this in a different branch, let me know what you guys prefer @shine2lay  @jusx

shine2lay:
> IMHO regex does not work well when the field is an array.
> it does work great when the field is a string, but not for an array

can you elaborate a bit more on this? I am not sure I understand fully

diegomarangoni:
> > IMHO regex does not work well when the field is an array.
> > it does work great when the field is a string, but not for an array
> 
> can you elaborate a bit more on this? I am not sure I understand fully

nevermind, I just saw how [changeset](https://mergeable.readthedocs.io/en/latest/validators/changeset.html) works, and now makes sense to me

diegomarangoni:
closing in favor of #496

PR #494, opened 2021-01-22
chore: update testing action node version to align with package.json

jusx:
## Changes
- chore: update testing workflow node version to align with package.json
- fix: the npm command test-coverage invokes an upload to codecov. Remove redundant GH action to upload which is a dup.
- fix: to invoke CI on push in a PR branch

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -5,7 +5,7 @@ on:
     branches: ['**']
   pull_request:
     branches: ['**']
-    types: [synchronize, opened]
+    types: [synchronize]
 
 jobs:
   build:
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/testing.yml
```diff
@@ -27,11 +24,3 @@ jobs:
     - run: npm run test-coverage
       env:
         CI: true
-    - name: Upload coverage to Codecov
```
shine2lay, line 30: i don't see the duplication part of this code (mentioned in the description). are we simply removing this?

shine2lay (commented):
left a comment

.github/workflows/testing.yml
jusx: The npm command includes an upload by calling codecov. This Github action does the same thing. In that respect, it's a duplication of task.

shine2lay (approved):

.github/workflows/testing.yml
shine2lay: ah okay, got it

PR #495, opened 2021-01-28
feat: global settings feature

iammuho:
**Problem**:
Since we are starting to enabling more feature toggles, we have introduced more environment variables, but these environment variables hard to manage and a bit messy. For any settings change, we need to re-deploy (for docker env) but for any environment setup, we need to restart the application.

**Solution:**
Like rule configuration file, I've introduced a `mergeable.settings.yml` file.

- If an organization-wide .github repo with mergeable.settings.yml then it should use it. (github.com/hellofresh/.github/.github/mergeable.settings.yml)
- if an organization wide settings do not exist, then it will use the one in path <owner>/<repo>/.github
(github.com/hellofresh/ahoy/.github/mergeable.settings.yml)

```
version: 1
mergeable:
  use_config_from_pull_request: true
  use_config_cache: false
  use_org_as_default_config: false
  config_path: 'mergeable.yml'
```

These options won't cause any problem if there is no settings file, because settings follow the below steps to define a setting;

- Check for the settings file
- Check for the environment variable
- Check for default


lib/settings/settings.js
```diff
@@ -0,0 +1,192 @@
...
+  NON_ARRAY_MERGEABLE: 80,
+  MISSING_WHEN_KEYWORD: 90,
+  MISSING_VALIDATE_KEYWORD: 100,
+  NON_ARRAY_VALIDATE: 110
+}
+Settings.ERROR_CODES = ERROR_CODES
+const ERROR_MESSAGES = {
+  MISSING_MERGEABLE_NODE: 'The `mergeable` node is missing.',
+  MISSING_RULE_SETS: '`mergeable` node does not contain any rule sets',
+  NON_ARRAY_MERGEABLE: '`mergeable` must be an array for version 2 config',
```
shine2lay, line 185: this should be `version 1 settings`

lib/settings/settings.js
```diff
@@ -0,0 +1,192 @@
...
+  MISSING_WHEN_KEYWORD: 'One or more rule set is missing `when` keyword',
```
shine2lay, line 186: i don't think this applies

lib/settings/settings.js
```diff
@@ -0,0 +1,192 @@
...
+  MISSING_VALIDATE_KEYWORD: 'One or more rule set is missing `validate` keyword',
```
shine2lay, line 187: i don't think this applies as well

lib/settings/settings.js
```diff
@@ -0,0 +1,192 @@
...
+  NON_ARRAY_VALIDATE: '`validate` must be an array of rules',
```
shine2lay, line 188: this doesn't apply either

lib/settings/settings.js
```diff
@@ -0,0 +1,192 @@
...
+  MISSING_RULE_SETS: 70,
+  NON_ARRAY_MERGEABLE: 80,
+  MISSING_WHEN_KEYWORD: 90,
+  MISSING_VALIDATE_KEYWORD: 100,
+  NON_ARRAY_VALIDATE: 110
+}
+Settings.ERROR_CODES = ERROR_CODES
+const ERROR_MESSAGES = {
+  MISSING_MERGEABLE_NODE: 'The `mergeable` node is missing.',
+  MISSING_RULE_SETS: '`mergeable` node does not contain any rule sets',
```
shine2lay, line 184: i think `rule sets` should be `settings`

lib/settings/settings.js
```diff
@@ -0,0 +1,192 @@
...
+  static getCache () {
+    return settingsCache
+  }
+
+  static instanceWithContext (context) {
+    return Settings.fetchSettingsFile(context).then(settings => {
+      return new Settings(settings)
+    }).catch(error => {
+      const settings = new Settings()
+      if (error instanceof SettingsFileNotFoundException) {
```
shine2lay, line 147: if the file is not found, shouldn't u just return default instead of throwing error? Since we are not expecting settings file to always be present

shine2lay (changes requested):
left a few comments

iammuho:
hey @shine2lay  I've fixed the requests, can you check that again ?

lib/settings/settings.js
shine2lay: I don't understand how this is being resolved, can you explain? i still see that the error is still set. 
Also it is not clear to me the behavior of the app when no settings file is found, can you help me understand?

iammuho:
hey @shine2lay , after merge this one, I'll configure the `cache manager` with the settings too. So waiting for you

lib/settings/settings.js
shine2lay: @iammuho can i get an answer for this? I have a feeling this will break production if deployed before it is going to expecting every repo to have settings file

iammuho: Hey @shine2lay ,

I've changed the code a bit for the cases if `mergeable.settings.yml` couldn't be found. Now default options will be passed to cache and context.

shine2lay (approved):

PR #496, opened 2021-01-28
feat: Filter validators

diegomarangoni:
As discussed in [#491](https://github.com/mergeability/mergeable/pull/491), this PR introduces a new feature called **Filters**.

With this feature, you can write conditions to run validators, for example:

```yaml
---
version: 2
mergeable:
    name: This only runs on private repos
    filter:
      - do: repository
        visibility: private
    validate:
      - do: approvals
        min:
          count: 1
          message: 'Developer review needed'
```

Or based on repository topics, like this:

```yaml
---
version: 2
mergeable:
    name: This only runs on public repos with library topic
    filter:
      - do: repository
        visibility: public
        topics:
          must_include:
            regex: '^library$'
    validate:
      - do: approvals
        min:
          count: 1
          message: 'Developer review needed'
        limit:
          teams: [ '@myorg/myteam' ]
```


lib/filters/options_processor/options/lib/constructOutput.js
```diff
@@ -0,0 +1,12 @@
+module.exports = (filter, input, rule, result, error) => {
```
shine2lay, line 1: I feel like there are alot of duplicate code (correct me if i am wrong), what do you think about refactoring the existing code to be reusable?

shine2lay (commented):
I noticed a few duplicate code, and it would also be nice to have tests and documentation for this feature

lib/filters/options_processor/options/lib/constructOutput.js
diegomarangoni: those functions are slightly different, as it's a `filter` and not a `validator`, I rename the object fields.
see if it's better now

diegomarangoni:
> I noticed a few duplicate code, and it would also be nice to have tests and documentation for this feature

working on tests now
can I do the documentation on a separate PR?

diegomarangoni:
I finished with tests, please review

samj1912:
How do these filters interact with schedule events? Do they work as expected?

jusx:
@samj1912 We've recently had to upgrade to the latest version of probot which probot-scheduler does not support. Furthermore https://github.com/probot/scheduler is no longer maintained. We have two options:

1. Deprecate the scheduler event.
2. Fork probot-scheduler and update it to work with the latest version of probot. This will incur maintenance cost.

Depending on what route we take, the scheduler events may no longer be relevant.

Would love to hear everyone's thoughts here. Is the scheduler event of value?

samj1912:
> @samj1912 We've recently had to upgrade to the latest version of probot which probot-scheduler does not support. Furthermore https://github.com/probot/scheduler is no longer maintained. We have two options:
> 
> 1. Deprecate the scheduler event.
> 2. Fork probot-scheduler and update it to work with the latest version of probot. This will incur maintenance cost.
> 
> Depending on what route we take, the scheduler events may no longer be relevant.
> 
> Would love to hear everyone's thoughts here. Is the scheduler event of value?
> 

It definitely is one of the most frequent use cases I use mergeable for. It's a shame to see it go. Are there any alternatives to probot scheduler?

jusx:
> > @samj1912 We've recently had to upgrade to the latest version of probot which probot-scheduler does not support. Furthermore https://github.com/probot/scheduler is no longer maintained. We have two options:
> > 
> > 1. Deprecate the scheduler event.
> > 2. Fork probot-scheduler and update it to work with the latest version of probot. This will incur maintenance cost.
> > 
> > Depending on what route we take, the scheduler events may no longer be relevant.
> > Would love to hear everyone's thoughts here. Is the scheduler event of value?
> 
> It definitely is one of the most frequent use cases I use mergeable for. It's a shame to see it go. Are there any alternatives to probot scheduler?

I don't know of an alternative. If we want to retain support for the scheduler event, I believe the best option is to fork it, fix the compatibility issues and continue maintenance from there.

samj1912:
> > > @samj1912 We've recently had to upgrade to the latest version of probot which probot-scheduler does not support. Furthermore https://github.com/probot/scheduler is no longer maintained. We have two options:
> > > 
> > > 1. Deprecate the scheduler event.
> > > 2. Fork probot-scheduler and update it to work with the latest version of probot. This will incur maintenance cost.
> > > 
> > > Depending on what route we take, the scheduler events may no longer be relevant.
> > > Would love to hear everyone's thoughts here. Is the scheduler event of value?
> > 
> > It definitely is one of the most frequent use cases I use mergeable for. It's a shame to see it go. Are there any alternatives to probot scheduler?
> 
> I don't know of an alternative. If we want to retain support for the scheduler event, I believe the best option is to fork it, fix the compatibility issues and continue maintenance from there.
> 
> 

Makes sense. Would you want to fork it, or simply add it to mergeable? The actual scheduler code is fairly small and simple and it could easily fit in a PR. I think once it works with mergeable, then we can generalize it and make a standalone module that others can use.

shine2lay (approved):
LGTM

shine2lay:
@samj1912 @jusx i've created a issue to track the conversation and progress on probot scheduler , #499

shine2lay:
@jusx , the `codecov/project` is failing, any clue as to why?

jusx:
> @jusx , the codecov/project is failing, any clue as to why?

@shine2lay this is an issue with the GH action which at the time was only being triggered for push and it does not seem to trigger on forked PR. 

I've added a trigger for the `pull_requests` event as well. The downside to this is that when we do this, it will be triggered twice when the PR is off master on the same repos.

In this PR the codecov is waiting for test results and never got it since the GH action isn't triggering.

shine2lay:
@jusx is there a way to re-trigger github action? or should we just merge it with out `codecov/project` check

jusx:
> @jusx is there a way to re-trigger github action? or should we just merge it with out codecov/project check

I updated the GH action to a catch all branch name. Updated this branch... seem to have triggered it now.

jusx:
@shine2lay And when the Codecov GH App is still waiting for status it's because the commit SHA is not matching some where. This seem to be a common problem (i.e https://community.codecov.io/t/coverage-status-on-commits-and-pull-requests-is-not-working/1148/5) -- I do not yet have a solution for this.

shine2lay:
@jusx should we go back to circle CI for now for reliability?

PR #497, opened 2021-01-29
docs: added options to docs

iammuho:
This PR adds docs for validator options.

Sample SS:

Screenshot 2021-01-29 at 11 06 05
Screenshot 2021-01-29 at 11 06 23

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,7 +1,6 @@
 CHANGELOG
 =====================================
 
-| January 29, 2021 : feat: added options to docs
 | January 21, 2021 : fix: request_review action failing when the reviewer is PR author `#486 <https://github.com/mergeability/mergeable/issues/486>`_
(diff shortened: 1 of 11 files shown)


jusx (commented):
This looks good. Just a suggestion that perhaps it will be helpful if under the "Supported Validators" where we list the validators we can link to the validators' respective docs. What do you think?

Will you be adding `jira` here?

iammuho:
Hey @jusx , yeah you just merged the JIRA so now I'll add the jira too. And I'll check linking!

iammuho:
Hey @jusx , I've looked at linking the highlighted sources part but I couldn't manage it and found out this plugin https://www.sphinx-doc.org/en/master/usage/extensions/viewcode.html which we might need to enable. Can you check this ?

shine2lay (approved):
LGTM

PR #503, opened 2021-02-07
fix: scheduler support

### LINKED ISSUE #499: Support for probot scheduler
We've recently had to upgrade to the latest version of probot which probot-scheduler does not support. Furthermore https://github.com/probot/scheduler is no longer maintained. We have two options:

Deprecate the scheduler event.
Fork probot-scheduler and update it to work with the latest version of probot. This will incur maintenance cost.
Depending on what route we take, the scheduler events may no longer be relevant.

Would love to hear everyone's thoughts here. Is the scheduler event of value?  (Taken from this [comment](https://github.com/mergeability/mergeable/pull/496#issuecomment-771267951))

please check the comment to see current discussion.

shine2lay:
The code used here is a fork of the https://github.com/probot/scheduler . 

closes #499


scheduler/index.js
```diff
@@ -0,0 +1,116 @@
...
+const ignoredAccounts = (process.env.IGNORED_ACCOUNTS || '')
+  .toLowerCase()
+  .split(',')
+
+const defaults = {
+  delay: !process.env.DISABLE_DELAY, // Should the first run be put on a random delay?
+  interval: 60 * 60 * 1000 // 1 hour
+}
+
+module.exports = (app, options) => {
```
jusx, line 13: I recommend commenting here on forking the code with the link to the original repos and why:

1. original version no longer maintained.
1. needed compatibility with the latest version of probot.

docs/changelog.rst
```diff
@@ -1,6 +1,7 @@
 CHANGELOG
 =====================================
 
+| February 6, 2021 : feat: Scheduler support `#499 <https://github.com/mergeability/mergeable/pull/499>`_
```
jusx, line 4: Don't necessarily view this as a new feature though. Just that we broke it with the upgrade of probot hence it's more of a fix.

shine2lay:
@jusx comment addressed PTAL

jusx (approved):

Issue #502, opened 2021-02-05
Central Cache


iammuho:
**Problem:**

Today we have encountered a problem with the configuration cache. There is a method called `resetCache` which clears the cache if PR has `mergeable.yml` but the problem is it clears in `memory` so `pod` level (for k8s users) but when you have more than 1 pods, some pods are fresh and some of them are being out-of-sync.

**Solution:**

To solve this problem, we need to create a feature to enable REDIS to support central cache.

samj1912:
Yeah, I had a similar issue. I fixed it by simply adding a small ttl of 30 seconds. Helps with bursty events and still works as intended.

shine2lay:
implemented in #506

PR #506, opened 2021-02-10
feat: global cache manager

iammuho:
Related issue: https://github.com/mergeability/mergeable/issues/502

--- 

To solve this issue and create functionality for the future, I've created a cache manager to be able to make the logic out of other controllers. With this feature, if anyone needs a cache functionality, they don't have to care about the store but just use the adapter functions. Cache manager will handle the rest depends on the ENV variables(or global settings future) for the CACHE_STORAGE var.

With this feature, we are introducing new ENVIRONMENT variables and we cannot place them to settings, because cacheManager works before settings file parsed. New environment variables are placed onto `.env.example` file

Feature list:

- [x] IN-Memory cache is working
- [x] Redis cache is working


lib/configuration/configuration.js
```diff
@@ -1,8 +1,9 @@
 const yaml = require('js-yaml')
 const _ = require('lodash')
-const NodeCache = require('node-cache')
+const CacheManager = require('../cache/cache')
 
-const configCache = new NodeCache({ stdTTL: 0, maxKeys: 5000 })
+// Setup the cache manager
+var cacheManager = new CacheManager()
```
shine2lay, line 6: can we please use either `const` or `let` instead of `var` for consistency sake?

shine2lay (approved):
LGTM, have one tiny suggested change.
Just food for thought, should we add this to `settings` file as well once it's been merged?

lib/cache/cache.js
```diff
@@ -0,0 +1,20 @@
+var cacheManager = require('cache-manager')
+
+class Cache {
+  constructor () {
+    this.cache = null
+
+    switch (process.env.CACHE_STORAGE) {
+      case 'redis':
+        var redisStore = require('cache-manager-ioredis')
+        this.cache = cacheManager.caching({ store: redisStore, url: process.env.REDIS_URL, refreshThreshold: 3 })
```
shine2lay, line 10: should `refreshThreshold` be a configurable variable?

lib/cache/cache.js
```diff
@@ -0,0 +1,20 @@
...
+        break
+      default:
+        this.cache = cacheManager.caching({ store: 'memory', max: 100, ttl: 10 })
```
shine2lay, line 13: what happens if there are more than 100 repos, what happens then? is it only storing the last 100 repo settings?

iammuho: hey @shine2lay,

Check the stackoverflow answer: https://stackoverflow.com/a/24230894

lib/cache/cache.js
iammuho: Hey you are right, I've moved every variable to environment variables. Defaults has been defined by the parent package already,

```
max: 500
ttl: null
```

iammuho:
> LGTM, have one tiny suggested change.
> Just food for thought, should we add this to `settings` file as well once it's been merged?

After you merge the settings pages, I'll modify here to use `global settings` for cache configuration.

iammuho:
Hey @shine2lay I've migrated settings to the `cacheManager` too and waiting for your reviews :)

jusx (approved):
Looks good. I like that your choice of switching to `cache-manager`

iammuho:
hey @jusx or @shine2lay , waiting for your merge. Fixed the conflicts

shine2lay:
LGTM, merging

PR #508, opened 2021-02-18
fix: minor bugs

shine2lay:
Includes following minor changes: 
- Make `review_request_fail_error` a log `info` instead of `error`
- add missing `actionObj` to deleteComment function
- skip checkRun if `ouptut` field is not present or `pull_request` is empty

### DIFF
--- a/lib/actions/request_review.js
+++ b/lib/actions/request_review.js
@@ -16,7 +16,7 @@ const createRequestReview = async (context, number, reviewers, actionObj) => {
       action_name: actionObj.name
     }
 
-    actionObj.log.error(JSON.stringify(errorLog))
+    actionObj.log.info(JSON.stringify(errorLog))
   }
   return res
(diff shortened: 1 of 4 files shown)


lib/actions/request_review.js
```diff
@@ -16,7 +16,7 @@ const createRequestReview = async (context, number, reviewers, actionObj) => {
       action_name: actionObj.name
     }
 
-    actionObj.log.error(JSON.stringify(errorLog))
+    actionObj.log.info(JSON.stringify(errorLog))
```
jusx, line 19: What is the reasoning for this change?

jusx (approved):

lib/actions/request_review.js
shine2lay: because we are using `info` for other known errors as well

Issue #509, opened 2021-02-20
option to use username list for approval.limit feature


vsingal-p:
Sometimes there is a need of adding a single username for the approval validation to a certain repo, whereas a common team is used across repos for restricting core approval to certain group of people.

Currently to achieve the same, we'll have to a create a repo specific team. It just adds an overhead for maintenance of multiple teams.

It would be great if we can have something like following - 
```
validate:
      - do: approvals
        min:
          count: 1
        limit:
          team: [ 'some-common-team-shared-across-repos' ]
          users: ['custom-user-1-for-this-specific-repo', 'custom-user-2-for-this-specific-repo']
```

shine2lay:
@vsingal-p I think that is a great use case!

shine2lay:
@vsingal-p you mentioned limiting the approval from user for only specific repos, would it make more sense to use the add a new `filter` that will only run the validations on when the repo is as specified? 

I was thinking of something like this: 
```
version: 2
mergeable:
  - when: pull_request.*
     filter: // this will check whether the validation should be run
       - do: repository
          name:
            must_include:
              regex: 'specific repo'
    validate:
      - do: approval
         min:
           count: 1
          limit:
            team: [ 'some-common-team-shared-across-repos' ]
            users: ['custom-user-1', 'custom-user-2']
```

vsingal-p:
The above would make sense only if we have a common mergeable.yml at the organisation level
Even in that case, we could be needing different set of users for different repos.

Currently we have a mergeable.yml in the repo where it's needed, so filter won't be of use to us

Thanks for sharing the idea though, the `or-required` is a good combination I could use in some of my use cases. Never looked it that way

vsingal-p:
And thanks for the quick turnaround adding adding the "users" feature 👯

vsingal-p:
One more question, right now if the following validations fail
```
validate:
      - do: approvals
        min:
          count: 1
        limit:
          team: [ 'some-common-team-shared-across-repos' ]
          users: ['user-1', 'user-2]
```

This is the error published in the checks report
```
❌ Validator: APPROVALS
     ❌ approvals count is less than "1"
     Input :
     Settings : {"min":{"count":1}}
```



**Can we make some change so that the following error is published?**
```
❌ Validator: APPROVALS
     ❌ approvals count is less than "1"
     Input :
     Settings : {"min":{"count":1}}

     ❌ Only approvals from following sources are counted.
     Input :
[...]

PR #510, opened 2021-02-22
feat: add `name` sub option to `repository` filter

shine2lay:
sample config
```
- do: repository
        name:
          must_include:
            regex: 'my-repo-name'
          must_exclude:
            regex: 'other-repo-name'
```

https://github.com/mergeability/mergeable/issues/509#issuecomment-782806518

### DIFF
--- a/__fixtures__/unit/helper.js
+++ b/__fixtures__/unit/helper.js
@@ -16,6 +16,7 @@ module.exports = {
         sha: 'sha1',
         action: 'opened',
         repository: {
+          name: (options.repoName) ? options.repoName : 'repoName',
           full_name: 'name',
           owner: {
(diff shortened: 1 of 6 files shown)


jusx (commented):
If this is a new feature, would it make sense to add user documentation as well?

shine2lay:
> If this is a new feature, would it make sense to add user documentation as well?

I've already added documentation to the PR

jusx (approved):
I am unaware of a specific use-case where this would be useful but code wise it looks fine.

shine2lay:
one use case that comes to mind where this could be useful is in org wide config file, they can have multiple generic recipes that applies to repos that `includes/excludes` specific names

PR #511, opened 2021-02-22
feat: add `limit.users` option to approval validator

### LINKED ISSUE #509: option to use username list for approval.limit feature

shine2lay:
closes #509

### DIFF
--- a/docs/validators/approval.rst
+++ b/docs/validators/approval.rst
@@ -23,6 +23,9 @@ Approvals
 
 
 .. note::
+    in ``limit`` options, if more than one sub option is present, the union of the results will be used.
+
+.. note::
     If you receive an error for `Resource not accessible by integration' for Owners file, it means you haven't given mergeable read file permission
 
 .. note::
--- a/docs/validators/approval.rst
+++ b/docs/validators/approval.rst
@@ -18,6 +18,7 @@ Approvals
         message: 'Custom message...'
       limit:
         teams: ['org/team_slug'] # when the option is present, only the approvals from the team members will count
+        users: ['user1', 'user2'] # when the option is present, approvals from users in this list will count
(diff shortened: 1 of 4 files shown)


vsingal-p (approved):

vsingal-p:
@jusx Can you please review this?

jusx (approved):
LGTM

docs/validators/approval.rst
```diff
@@ -18,6 +18,7 @@ Approvals
         message: 'Custom message...'
       limit:
         teams: ['org/team_slug'] # when the option is present, only the approvals from the team members will count
+        users: ['user1', 'user2'] # when the option is present, approvals from users in this list will count
```
jusx, line 21: I'm not sure this comment is clear:

1. Does it mean that only these users listed will be summed as the total required approval count?
1. Does it mean that only this subset of users from the team members will count?

I believe the documentation needs work and a comment in this example config is not enough.

I suggest a more detailed description outside of a settings sample.

vsingal-p: It's option 1 which you mentioned

+1, the documentation can be improved

shine2lay: added a note on the page to try and clarify it.

PR #512, opened 2021-02-22
feat: add a info panel for 'limit' option in `approval` validator

shine2lay:
see the discussion in https://github.com/mergeability/mergeable/issues/509#issuecomment-783346365

output
image

### DIFF
--- a/lib/validators/approvals.js
+++ b/lib/validators/approvals.js
@@ -101,7 +101,7 @@ class Approvals extends Validator {
       teamMembers = _.union(teamMembers, owners)
       let validatorContext = { name: 'approvals' }
 
-      output.push(constructOutput(validatorContext, _.union(teams, owners), limitOption, {status: 'info', description: `Only the following teams and users approval will counted`}))
+      output.push(constructOutput(validatorContext, approvedReviewers, limitOption, {status: 'info', description: `Only approvals from following sources are counted`}))
       approvedReviewers = _.intersection(approvedReviewers, teamMembers)
     }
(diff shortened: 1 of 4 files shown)


lib/validators/approvals.js
```diff
@@ -98,11 +99,14 @@ class Approvals extends Validator {
       }
 
       teamMembers = _.union(teamMembers, owners)
+      let validatorContext = { name: 'approvals' }
 
+      output.push(constructOutput(validatorContext, _.union(teams, owners), limitOption, {status: 'info', description: `Only the following teams and users approval will counted`}))
```
vsingal-p, line 104: You can use "Only approvals from following sources are counted." verbiage instead
This will cover all future options in `limit` as well

vsingal-p: Should we use `_.union(teams, owners)` or `approvedReviewers` ?

shine2lay: i thought input should be `sources` rather than the current `approvedReviewers` that is counted

shine2lay: do you think `approvedReviewers` make more sense here??

vsingal-p: Yeah, because `input` is already displayed in `settings` field in the output

shine2lay:
@jusx can you please review this?

vsingal-p:
@jusx Can you please review this?

jusx (approved):
The informational tip LGTM

PR #513, opened 2021-02-22
docs: add reusabe config section

### LINKED ISSUE #478: Documentation for re-usable validation rules or actions
There are cases where there is a need to repeat validation or action rules in different recipes. Using YAML advanced features [called anchors](https://yaml.org/spec/1.2/spec.html#id2785586), we can avoid configuration duplication.

This is a little known feature in YAML yet can be very useful when dealing with multiple recipes or large recipes. We should document this and show examples on how it can be used with a IRL use case.

shine2lay:
closes #478

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -153,7 +153,7 @@ Actions that mergeable is currently able to perform.
 
 .. _config-reuse-page:
 
-Reusable Config
+Reusable Configuration
 --------------------------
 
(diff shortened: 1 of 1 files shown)


docs/configuration.rst
```diff
@@ -151,6 +151,39 @@ Actions that mergeable is currently able to perform.
     actions/labels.rst
     actions/request_review.rst
 
+.. _config-reuse-page:
+
+Reusable Config
```
jusx, line 156: I recommend spelling out `Configuration`

shine2lay: addressed

jusx (approved):

PR #519, opened 2021-02-27
WIP: test error logger/catcher layer

shine2lay:
Currently, we have a catch all errors for `validators, filters and actions` but it can be very hard to debug since it's hard to pin point which `validator, filter or action` threw the error. 
This is an attempt to address it by catching all the github api call error and logging them.
I also noticed that there are multiple calls to the same end point from different location, this could help consolidate those

NOTE: this is just a sample for discussion purposes, not the complete implementation of the feature


jusx (approved):
abstraction looks good to me.

shine2lay:
closed, a complete version is at #524

PR #524, opened 2021-03-13
refactor: githubAPI abstraction layer

shine2lay:
Refactor github api calls into an abstraction layer for two purposes:
- reusablility
- to catch and log errors 

Note: this PR only include one particular github API call, in order to keep the PR small. other github calls will be refactor in subsequent PRs


shine2lay:
@jusx In this PR, I've set the default behaviour when an error is found is to just log it and re throw it (creating the same behaviour as what we have now). Do you have any other suggestions on the behaviour?

jusx (approved):

PR #563, opened 2021-07-08
docs: update supported events for merge action

tenshiAMD:
As I'm reading the documentation, I noticed that the support events in code do not match in the documentation (https://github.com/mergeability/mergeable/blob/master/lib/actions/merge.js#L20:L24).


shine2lay:
thanks for catching the mistake and fixing it

shine2lay (approved):

Issue #550, opened 2021-06-10
No New Typescript Files: Request for Advice


jaisal1024:
I am trying to implement a rule that will disallow introducing new javascript files into our codebase. So one should be allowed to add new code to a current javascript file, but disallowed from creating a new javascript file in our code base. 

I looked at using the content and changeset validators but neither provides a way to include only created files and not just modified files. 

I am proposing adding a new flag to content. Something like `createdFilesOnly`. What are the authors thoughts on this use case and solution?

If this works out, I would also love to add the example for others to see :)

shine2lay:
@jaisal1024 hey, sorry for the late reply. 

Yeah, I don't think we currently have a way to only check changeset against added files.
It can be easily added however, you just need to change this one file https://github.com/mergeability/mergeable/blob/master/lib/validators/changeset.js

as for settings, I think
```
files:
  added: true // allow added files in the changeset against validation
  modified: true // allow modified files in the changeset against validation
  deleted: true //  allow deleted files in the changeset against validation
```

I think with this setting, we can mix and match which type of files to filter in/out.

I am busy these days, I'll try to pick up when I have time. If you want to create a PR, I'll do my best to review and merge it in ASAP

jaisal1024:
We will open a PR for this soon!

PR #564, opened 2021-07-12
feat: Filter specific status of files in changeset

gorosgobe:
This PR implements the changes described in https://github.com/mergeability/mergeable/issues/550

Thanks to Bloomberg L.P. for letting us contribute to Mergeable.

### DIFF
--- a/docs/validators/changeset.rst
+++ b/docs/validators/changeset.rst
@@ -25,6 +25,10 @@ Change set
       max:
          count: 2 # max number of files in a PR
          message: 'Custom message...'
+      files:
+         added: true
+         modified: false # setting to false is optional
+         deleted: true
       # all of the message sub-option is optional
 
 you can use ``and`` and ``or`` options to create more complex validations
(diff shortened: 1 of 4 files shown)


docs/validators/changeset.rst
```diff
@@ -25,6 +25,10 @@ Change set
       max:
          count: 2 # max number of files in a PR
          message: 'Custom message...'
+      files:
+         added: true
```
shine2lay, line 29: for consistency sake, would you please add the `default` value of each field? i.e
`added: true #default : true, if false, added files will not counted toward changeset`

docs/validators/changeset.rst
```diff
@@ -25,6 +25,10 @@ Change set
       max:
          count: 2 # max number of files in a PR
          message: 'Custom message...'
+      files:
```
shine2lay, line 28: It would be helpful if you can add an comment about what this setting is for. just for clarity sake.

shine2lay (approved):
LGTM! a few minor comments on the docs section.

Issue #565, opened 2021-07-12
Preventing validators to run on `draft` PRs.


hemeroc:
If I understand this PR https://github.com/mergeability/mergeable/pull/515 correct the new payload filter enables a user to create a filter for any payload element.

**Expectation:**

> Therefore my expectation would be that I can use that filter to run specific validators not on `draft` PRs.
> As far as I have seen that does not work due to the fact that the filter does not support boolean/integer checks.

**Reason:**

>I would need this feature mainly to prevent `stale` checks from running on `draft` PRs.

Please let me know if my assumptions are correct or if I am thinking in a wrong direction.
Other options would be:
- adding a validator for `darft` PRs
- making this an option of the `stale` validator

Also let me know if I can support you adding this possibility.

shine2lay:
@hemeroc I think you are right, the only thing preventing this use case is the lack of checks against `boolean` in `payload` filter.

I think you'll need to add the `boolean` option to this line https://github.com/mergeability/mergeable/blob/master/lib/filters/payload.js#L7 
and add something similar to this option processor https://github.com/mergeability/mergeable/blob/master/lib/validators/options_processor/options/no_empty.js 
within the `lib/filters/options` folder

I won't be able to get to this anytime soon, if you do create a PR however, i'll make sure it gets merged ASAP

hemeroc:
For ease of implementation I went for implementing an option on the the stale validator called `ignore_drafts`.
Similat to the project and milestone options.

shine2lay:
@bbugh yes, I was recommending something like that, to be consistent with current implementation. do you have a recommendation?

bbugh:
No better recommendations. :) Now that I'm clear on what you were proposing, I will submit a PR for this.

shine2lay:
@bbugh awesome! 
What do you think about using `on` instead of `enabled`? I think `enable` is not as intuitive if you just look at the config. 

- do: payload
        pull_request:
          draft:
            boolean:
              # don't run on `draft: true` (since this is a filter, only if the condition is true, the validation should would run)
              on: false

bbugh:
I was playing with `match` like some of the other options, since we want to validate that it's an exact value. There's also `is`. I think the equality synonym is useful. Like:

```
      - do: payload
        pull_request:
          draft:
            boolean:
              # like current options
              match: false
			  # or
              is: false
```

shine2lay:
[...]

PR #567, opened 2021-07-19
feat: Add `ignore_drafts` option to ignore drafts in stale validator

hemeroc:
See https://github.com/mergeability/mergeable/issues/565

This is a very small fix to ignore drafts in the stale validator.

### DIFF
--- a/__tests__/unit/validators/stale.test.js
+++ b/__tests__/unit/validators/stale.test.js
@@ -55,8 +55,17 @@ describe('metadata queries', () => {
 
   })
 
-  test('will set milestones correctly', async () => {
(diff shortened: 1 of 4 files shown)


__tests__/unit/validators/stale.test.js
```diff
@@ -55,8 +55,17 @@ describe('metadata queries', () => {
 
   })
 
-  test('will set milestones correctly', async () => {
-    const settings = { do: 'stale', days: 10, ignore_milestones: true }
+  test.each([
```
hemeroc, line 58: As I was adding another dimension here I created a parameterized test instead

shine2lay (approved):
LGTM!

PR #568, opened 2021-07-26
chore: Add npm publish workflow

shine2lay:
Every time master branch is `pushed`, if the version is different from current, it'll publish the npm package


shine2lay:
closed becase a newer PR of this #569 is opened

Issue #573, opened 2021-08-03
Allow array of regexes (or strings) for must_include, must_exclude


bbugh:
Hi! 👋 This is a great tool!

I've run into a snag, though. I have a list of about 30 exact filenames that need to be matched in a monorepo (configuration, build related things), and doing it all as one regex is very difficult for readability and writability.

https://github.com/mergeability/mergeable/blob/01fd21a8f81e303ee30d660452047601ba2adf3f/lib/validators/contents.js#L21-L30

Changing these to allow an array in `regex` would be really helpful. Adding `match` like `begins_with` and `ends_with` that does an exact match would be even better to avoid accidental matches or having to mark every filename with boundaries.

Thanks!

shine2lay:
@bbugh Hey! 

re `match`: for `begins_with` and `ends_with`, we just convert them to regex to match anyways, I see your point about potentially troublesome regex.

Have you try `and` or `or` option? I think they could be useful for your use case. 
something like this, they can also be nested

```
- do: contents
  and:
    - must_include:
        regex: 'doc/.*'
        message: 'Custom message...'
    - must_include:
        regex: 'changelog.md'
        message: 'Custom message...'
  or:
    - must_include:
        regex: 'package-lock.json'
        message: 'Custom message...'
    - must_include:
        regex: 'yarn.lock'
        message: 'Custom message...'
```

bbugh:
Yeah, that was what prompted me to write this. Written out with `or`, that one validation is 99 repetitive lines long. 😬 Having a shorter way to write it like the other matchers would be 🔥

shine2lay:
@bbugh oh, I see. In that case, I don't think I'll get around to this anytime soon, if you want to create a PR for this, I'll do my best to get it merged asap.

I think you just need to change code in this file to support `must_include array` for all validators, you'll need to add 'array' to supported settings too https://github.com/mergeability/mergeable/blob/01fd21a8f81e303ee30d660452047601ba2adf3f/lib/validators/options_processor/options/must_include.js

bbugh:
@shine2lay thanks for the quick reply, happy to PR. 

It appears that `MustInclude` already supports an array type, am I reading this correctly?

https://github.com/mergeability/mergeable/blob/01fd21a8f81e303ee30d660452047601ba2adf3f/lib/validators/options_processor/options/must_include.js#L31-L41

[...]

PR #575, opened 2021-08-05
feat: Support must_include and must_exclude regex as an array

bbugh:
Hi! :wave: Another PR. This solves our use case raised in #573 and allows `must_include` to be either a `string` or an `array.

I imagine this would be used in other places, but I'm not sure where.

### DIFF
--- a/__tests__/unit/validators/changeset.test.js
+++ b/__tests__/unit/validators/changeset.test.js
@@ -18,6 +18,29 @@ test('validate returns correctly', async () => {
   expect(validation.status).toBe('fail')
 })
 
+test('supports arrays', async () => {
+  const changeset = new Changeset()
+  const settings = {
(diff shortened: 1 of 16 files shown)


shine2lay (approved):
Looks good, just for symmetry sake, would you mind doing the same for `must_exclude` as well? I think you also need to update the docs and changelog

bbugh:
Alright, I tested locally and it's working great. I extended it to include `must_exclude`, added some more tests, and updated the other validators that use `must_include` and `must_exclude` to support the array format. Also added the changelog. Let me know if there's anything else you want me to address!

bbugh:
Oh, also updated the docs for the `MustInclude` and `MustExclude` pages, though I'm not sure if I got everything.

shine2lay (approved):

PR #577, opened 2021-08-10
feat: New labels API 

### LINKED ISSUE #576: Adding and removing labels in the same pass results in removing the added label
Adding and removing labels in one action results in the newly added label being removed immediately, I believe because `deleteLabels` is looking for the issue's labels, but the operations run asynchronously so it's a race condition. 

image

We cannot use `replace` because we have other labels that need to stay. 

## Potential fixes

I'm happy to PR this, and here's a couple of possible options to resolve it. Do you have a preference?

1. If `replace` supported a "replace these labels with these other labels" rather than removing them all, we could use it.
	```yaml
	  - do: labels
	    labels: ["triaged"]
        replace: ["needs triage"]
        mode: "replace"
    ```

2. Change the `labels` action to allow multiple operations at once, so that the app can get a safe list in one batch. This could get the full list of labels first, make adjustments, and then use `setLabels` for the result.
    ```yaml
      - do: labels
        add: ["needs review"]
        delete: ["work in progress"]

      # or: 

      - do: labels
        replace: ["triaged"]
    ```

I think \#2 is best as it has a consistent API, and the `labels` and `type` options could be cleanly deprecated with a warning.

## Reproduction steps

Given this:

```yaml
    pass:
      - do: labels
        labels: ["needs review"]
        mode: "add"
      - do: labels
        labels: ["work in progress"]
        mode: "delete"
```

The result is this, if `work in progress` is not labeled:

image

If it is labeled `work in progress`, it does this:

image

Changing the order of the `labels` operation in the configuration does not change the outcome.

bbugh:
This adds the `add`, `replace`, `delete` options for the `labels` action. Documentation was extended to include the new info, and labeled the old API as deprecated. I tested a bunch of scenarios with this with a local deploy and it's working great!

Closes #576

### DIFF
--- a/__tests__/unit/actions/labels.test.js
+++ b/__tests__/unit/actions/labels.test.js
@@ -2,6 +2,76 @@ const Labels = require('../../../lib/actions/labels')
 const Helper = require('../../../__fixtures__/unit/helper')
 const UnSupportedSettingError = require('../../../lib/errors/unSupportedSettingError')
 
+test('check that replace replaces existing labels', async () => {
+  const labels = new Labels()
+  const context = createMockContext(['drop_label'])
+
+  const settings = {
+    replace: ['work in progress', 'development']
+  }
+
+  await labels.afterValidate(context, settings)
+  expect(context.octokit.issues.setLabels.mock.calls.length).toBe(1)
+  expect(context.octokit.issues.setLabels.mock.calls[0][0].labels).toStrictEqual({ labels: settings.replace })
+})
+
+test('check that add appends to existing labels', async () => {
+  const labels = new Labels()
+  const context = createMockContext(['another label', 'test label'])
+
+  const settings = {
+    add: ['production', 'deploy']
+  }
+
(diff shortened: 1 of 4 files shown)


bbugh:
I'm not sure about the failing Code Coverage check. Running either `test` or `test-coverage` locally is successful:

```
Test Suites: 63 passed, 63 total
Tests:       499 passed, 499 total
Snapshots:   0 total
Time:        11.496 s
Ran all test suites.
```

bbugh:
@shine2lay Hey there! Is there anything else I can do to help get this merged? I'm not sure what is happening with the coverage, the report above says it only increased and has 100% coverage.

shine2lay:
@bbugh hey, thanks for ping, I must've missed it. I'll re run the CI and see if it was just a fluke.

shine2lay (approved):
LGTM!

bbugh:
Thanks so much! I really appreciate how on top of it you are with this project.

PR #583, opened 2021-08-24
Add boolean filter for payload

### LINKED ISSUE #565: Preventing validators to run on `draft` PRs.

bbugh:
This adds a `Boolean` filter option for `payload` that can validate whether an attribute or subtree attribute is exactly `true` or `false`.

This also fixes an issue where the link [the Payload docs](https://mergeable.readthedocs.io/en/latest/filters/payload.html) was not displaying correctly, which made most of the rest of the page not appear. 

- [ ] Add tests if behavior is correct

Closes #565

### DIFF
--- a/lib/validators/options_processor/options/boolean.js
+++ b/lib/validators/options_processor/options/boolean.js
@@ -1,0 +1,32 @@
+const MATCH_NOT_FOUND_ERROR = 'Failed to run the test because \'match\' is not provided for \'boolean\' option. Please check README for more information about configuration'
+const UNKNOWN_INPUT_TYPE_ERROR = 'Input type invalid, expected boolean as input'
+
+class BooleanOption {
+  static process (validatorContext, input, rule) {
+    const filter = rule.boolean
+
+    const match = filter.match
+    let description = filter.message
+    if (match == null) {
+      throw new Error(MATCH_NOT_FOUND_ERROR)
+    }
+
+    let isMergeable
+
+    const DEFAULT_SUCCESS_MESSAGE = `The ${validatorContext.name} is ${match}`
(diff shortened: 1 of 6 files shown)


lib/validators/options_processor/options/boolean.js
```diff
@@ -0,0 +1,32 @@
+const MATCH_NOT_FOUND_ERROR = 'Failed to run the test because \'match\' is not provided for \'boolean\' option. Please check README for more information about configuration'
+const UNKNOWN_INPUT_TYPE_ERROR = 'Input type invalid, expected strings "true" or "false", or boolean literal `true` or `false` as input'
+
+class BooleanMatch {
```
shine2lay, line 4: Since the `boolean` option is added under `validators`, could you please write a test case in one of the validator using `boolean`? just to ensure that the `boolean` option also work for `validators`, not just `payload`

shine2lay (changes requested):
looks great, just a small request to add tests for `validator` using `boolean` option

PR #585, opened 2021-08-26
fix: Await GithubAPI topics response

Aj-vrod:
Small fix for GithubAPI response


shine2lay (approved):
LGTM! thanks for the fix

PR #589, opened 2021-09-24
feat: add support for objects in arrays in the payload filter

jaredhobbs:
This adds an additional `key` attribute to the `must_include` and `must_exclude` options processors.

Use case for this was for the `payload` filter. I needed to filter on particular labels. The `labels` array returns objects so I added a `key` attribute to be able to filter on a field in the object.


shine2lay:
@jaredhobbs could you please add documentation and an entry to changelog, thanks.

lib/validators/options_processor/options/must_exclude.js
```diff
@@ -34,9 +35,23 @@ class MustExclude {
         return !regexObj.test(input)
       } else if (Array.isArray(input)) {
         if (filter.all) {
-          return input.every(label => !regexObj.test(label))
+          return input.every(label => {
+            if (typeof label === 'string') {
```
shine2lay, line 39: Not a big deal but I noticed this part of the code is abstracted out to `arrayHandler` in `must_include.js`, is there a reason it is not done here?

lib/validators/options_processor/options/must_exclude.js
```diff
@@ -34,9 +35,23 @@ class MustExclude {
...
+              return !regexObj.test(label)
+            } else if (key) {
+              return !regexObj.test(label[key])
+            }
+            return false
```
shine2lay, line 44: should this return an error instead? If the code reaches here, doesn't this mean that user has probably set some settings wrong?

shine2lay (changes requested):
Just have one question about whether to `return false` or `throw error`. the rest looks good.

lib/validators/options_processor/options/must_exclude.js
jaredhobbs: mainly because in the must_exclude case, the logic in filter.all and the else statement are slightly different:

`input.every(label => !regexObj.test(label))`

vs

`!input.some(label => regexObj.test(label))`

notice the position of the `!`

lib/validators/options_processor/options/must_exclude.js
jaredhobbs: good point, adding now

shine2lay (approved):
LGTM

PR #591, opened 2021-09-27
fix: use node version 14.17.6 for the release action

jaredhobbs:
The latest version of semantic-release requires node >=14.17 (https://github.com/semantic-release/semantic-release/pull/2132)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


jaredhobbs:
@shine2lay that last PR failed to release due to a change in the latest semantic-release. They now require node min version of 14.17. This PR updates the node version in the GitHub action that releases new versions.

shine2lay (approved):

PR #593, opened 2021-10-15
feat: Add boolean option for payload filter

### LINKED ISSUE #565: Preventing validators to run on `draft` PRs.

hemeroc:
This adds a Boolean filter option for payload that can validate whether an attribute or subtree attribute is exactly true or false.

This also fixes an issue where the link the Payload docs was not displaying correctly, which made most of the rest of the page not appear.

Closes #565 & #583

This PR is a duplicate of #583 that includes the [review feedback.](https://github.com/mergeability/mergeable/pull/583#discussion_r727427241)


shine2lay:
@hemeroc hey, thanks for the ping, sorry I dropped the ball on this

PR #612, opened 2022-02-03
feat: Add an option to use PR title as commit message

kichik:
Added a merge setting options called `commit_title` and `commit_message` that allow the user to control the merged commit message. These are templates that use the PR object as variables, so you can use the PR title, body, and number, for example. This allows users to set rules requiring certain commit messages without forcing the user to squash and `git push --force`. If we want the commit message to always contain the ticket, just force the user to update the PR title instead of pushing again.

### DIFF
--- a/lib/actions/merge.js
+++ b/lib/actions/merge.js
@@ -11,11 +11,11 @@ const mergePR = async (context, prNumber, mergeMethod, commitTitle, commitMessag
 
     if (pullRequest.data.mergeable_state !== 'blocked' && pullRequest.data.state === 'open') {
       const mergeParams = { pull_number: prNumber, merge_method: mergeMethod }
-      if (commitTitle) {
+      if (commitTitle !== undefined) {
         const template = handlebars.compile(commitTitle)
         mergeParams.commit_title = template(pullRequest.data)
       }
-      if (commitMessage) {
+      if (commitMessage !== undefined) {
         const template = handlebars.compile(commitMessage)
         mergeParams.commit_message = template(pullRequest.data)
       }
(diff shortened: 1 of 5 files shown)


kichik:
It's working! I tested by running it locally with smee. I installed it to my fork and created the following:

* PR: https://github.com/kichik/mergeable/pull/3
* PR commit: https://github.com/kichik/mergeable/pull/3/commits/9a81a4cdd3fae47b834f6135ffa5344cb36bba5e
* Merge commit: https://github.com/kichik/mergeable/commit/75e5cd85430304db277dfc8499ff0cf3447fa3c6

MJCaddell (approved):

shine2lay:
i'll take a closer look at it after work today, please bear with me here

lib/actions/merge.js
```diff
@@ -37,6 +47,8 @@ class Merge extends Action {
       throw new UnSupportedSettingError(`Unknown Merge method, supported options are ${MERGE_METHOD_OPTIONS.join(', ')}`)
     }
     const mergeMethod = settings.merge_method ? settings.merge_method : 'merge'
+    const commitTitle = settings.commit_title ? settings.commit_title : ''
```
shine2lay, line 50: if you default it like this, wouldn't the default commit title be an empty string? same with commit message?
have you tested out the behavior when the template is not present in settings? I think it's important for backward compatibility

shine2lay (commented):
Have one question

lib/actions/merge.js
kichik: I added a test case to confirm this works as expected when the user doesn't set title or message, and it does. But what this also means is that the user can't set an empty message in YAML. So I went ahead and also updated the code to support empty strings. Thanks for catching this!

shine2lay (approved):

shine2lay:
@kichik LGTM! can you please add a changelog entry? then i can merge it. Thanks

shine2lay (approved):

PR #613, opened 2022-02-04
fix: Support running on Windows with cross-env

kichik:


shine2lay:
maybe a dumb question, adding this doesn't affect current `linux` users right?

kichik:
It shouldn't. `cross-env` is the most common way to support both. It's used everywhere. Even [bootstrap](https://github.com/twbs/bootstrap/blob/main/package.json) uses it.

shine2lay:
okay, just double checking because I haven't used it personally

shine2lay (approved):

shine2lay (approved):

PR #623, opened 2022-02-24
fix: correct indentation on documentation

ChrisRBe:
Section is not shown as code but as commented text.


shine2lay (approved):
thanks a bunch

PR #626, opened 2022-03-22
fix: pass comment instance to removeErrorComments

diegomarangoni:
When an error comment is created for any reason (for example GitHub unstable) and then the error is fixed, then the Action cannot delete this comment because the invocation is wrong.

Call to `comment.removeErrorComments` must also send instance of `comment` as 2nd argument.


diegomarangoni:
hey @shine2lay, can I have your review?

diegomarangoni:
hey @shine2lay, can I have your review?

shine2lay (approved):

shine2lay:
thanks @diegomarangoni

PR #636, opened 2022-05-02
chore(fix): Switch to networking.k8s.io/v1

gferon:
`networking.k8s.io/v1beta1` was promoted to `networking.k8s.io/v1` in Kubernetes 1.19, and removed in Kubernetes 1.22.


gferon:
@shine2lay is there any chance you could give this a look? thanks!

Issue #654, opened 2022-08-24
An error occurs using `comment` action


tenshiAMD:
Greetings. I have a working configuration before and I found out that the `comment` action is not working anymore. Here's what the check details look like.

```
Some or All of the validators have returned 'error' status, please check below for details
  
  Here are some stats of the run: 
  1 validations were ran. 
  0 ***PASSED***
  0 ***FAILED***
  1 ***ERRORED***
```

tenshiAMD:
@shine2lay please check this [status check](https://github.com/ruby-gh-actions/sandbox-mergeable/pull/2/checks?check_run_id=8003279379) if you have time. Thanks in advance.

shine2lay:
@tenshiAMD I get 404 when I click the link

tenshiAMD:
@shine2lay looks like it needs permission, would you like to see it? I will need to add you as a collaborator to the repo then.

shine2lay:
@tenshiAMD sure, if you can add me. If not, a screenshot could work as well.

tenshiAMD:
@shine2lay Here's a screenshot, please check. Thanks!

<img width="1728" alt="comment account issue" src="https://user-images.githubusercontent.com/13580338/186527773-9062353d-a0e8-4d44-b417-aa7ed3e495ad.png">

shine2lay:
@tenshiAMD first thought is that the PR's description is somehow "undefined" instead of an empty string because the error is complaining about input.

tenshiAMD:
@shine2lay it is working before. My setup is we a dev missed an adding description, mergeable should leave a comment that the dev should add some description. I wonder it suddenly does not work anymore

tenshiAMD:
@shine2lay also status should show it as fail not as error

shine2lay:
@tenshiAMD I would be interested in checking the payload to see if `description` field is just an empty string (which is what is expected) or are we getting something else. It is causing error because `description` field is not a string (which is expected by the code). We definitely haven't done any changes that should impact this feature.

tenshiAMD:
@shine2lay or maybe the payload before is an empty string not it returns undefined or null. I'm pretty sure it is working fine before.

tenshiAMD:
@shine2lay hi I retraced the issue and found out that it all started here, I will add changes to check for input values (`null` and `undefined`)

[...]

PR #655, opened 2022-08-24
fix: set fallback value for `description` payload

tenshiAMD:
In case, the payload `body` somehow returns undefined or null, we set the fallback value to an empty string.

### DIFF
--- a/__tests__/unit/validators/description.test.js
+++ b/__tests__/unit/validators/description.test.js
@@ -28,6 +28,12 @@ test('isMergeable is false if the PR body is empty', async () => {
   let descriptionValidation = await description.processValidate(createMockPR(''), settings)
   expect(descriptionValidation.status).toBe('fail')
 
+  descriptionValidation = await description.processValidate(createMockPR(undefined), settings)
+  expect(descriptionValidation.status).toBe('fail')
+
+  descriptionValidation = await description.processValidate(createMockPR(null), settings)
+  expect(descriptionValidation.status).toBe('fail')
+
   descriptionValidation = await description.processValidate(createMockPR('Some Description'), settings)
   expect(descriptionValidation.status).toBe('pass')
 })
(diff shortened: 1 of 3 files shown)


tenshiAMD:
@shine2lay please check if this could be a fix for https://github.com/mergeability/mergeable/issues/654. Thanks in advance.

__tests__/unit/validators/description.test.js
```diff
@@ -30,6 +30,12 @@ test('isMergeable is false if the PR body is empty', async () => {
 
   descriptionValidation = await description.processValidate(createMockPR('Some Description'), settings)
   expect(descriptionValidation.status).toBe('pass')
+  
+  descriptionValidation = await description.processValidate(createMockPR(undefined), settings)
+  expect(descriptionValidation.status).toBe('pass')
```
shine2lay, line 35: shouldn't this be 'fail'? since `no_empty` is `enabled: true`

__tests__/unit/validators/description.test.js
```diff
@@ -30,6 +30,12 @@ test('isMergeable is false if the PR body is empty', async () => {
...
+  
+  descriptionValidation = await description.processValidate(createMockPR(null), settings)
+  expect(descriptionValidation.status).toBe('pass')
```
shine2lay, line 38: same as above

__tests__/unit/validators/description.test.js
tenshiAMD: @shine2lay yup, my bad. I changed it. Please check. Thanks!

__tests__/unit/validators/description.test.js
tenshiAMD: @shine2lay yup, my bad. I changed it. Please check. Thanks!

shine2lay (approved):

PR #657, opened 2022-08-25
fix: set fallback for `no_empty` options processor

tenshiAMD:
This is to set fallback for `no_empty` options processor.

### DIFF
--- a/lib/validators/options_processor/options/no_empty.js
+++ b/lib/validators/options_processor/options/no_empty.js
@@ -1,7 +1,7 @@
 const _ = require('lodash')
 
 const ENABLED_NOT_FOUND_ERROR = 'Failed to run the test because \'enabled\' is not provided for \'no_empty\' option. Please check README for more information about configuration'
-const UNKNOWN_INPUT_TYPE_ERROR = 'Input type invalid, expected string as input'
+const UNKNOWN_INPUT_TYPE_ERROR = 'Input type invalid, expected string or Array as input'
 
 class NoEmpty {
   static process (validatorContext, input, rule) {
--- a/lib/validators/options_processor/options/no_empty.js
+++ b/lib/validators/options_processor/options/no_empty.js
@@ -1,3 +1,5 @@
+const _ = require('lodash')
+
 const ENABLED_NOT_FOUND_ERROR = 'Failed to run the test because \'enabled\' is not provided for \'no_empty\' option. Please check README for more information about configuration'
 const UNKNOWN_INPUT_TYPE_ERROR = 'Input type invalid, expected string as input'
 
@@ -18,7 +20,7 @@ class NoEmpty {
       }
     }
 
-    let isMergeable
+    let isMergeable = false
 
     const DEFAULT_SUCCESS_MESSAGE = `The ${validatorContext.name} is not empty`
     if (!description) description = `The ${validatorContext.name} can't be empty`
@@ -27,7 +29,7 @@ class NoEmpty {
       isMergeable = input.trim().length !== 0
     } else if (Array.isArray(input)) {
       isMergeable = input.length !== 0
-    } else {
+    } else if (!(input == null || _.isUndefined(input))) {
       throw new Error(UNKNOWN_INPUT_TYPE_ERROR)
     }
 
--- a/__tests__/unit/validators/options_processor/options/no_empty.test.js
+++ b/__tests__/unit/validators/options_processor/options/no_empty.test.js
@@ -31,6 +31,8 @@ test('return pass if input meets the criteria', () => {
 
 test('return fail if input does not meet the criteria', () => {
   verify(true, '', [], 'fail')
+  verify(true, null, [], 'fail')
+  verify(true, undefined, [], 'fail')
   verify(false, '', [''], 'pass')
 })
 
--- a/__tests__/unit/validators/options_processor/options/no_empty.test.js
+++ b/__tests__/unit/validators/options_processor/options/no_empty.test.js
@@ -43,7 +43,7 @@ test('return error if input does not meet the criteria', () => {
     const config = noEmpty.process(validatorContext, input, rule)
     expect(config).toBeDefined()
   } catch (e) {
-    expect(e.message).toBe('Input type invalid, expected string as input')
+    expect(e.message).toBe('Input type invalid, expected string or Array as input')
   }
 
   input = [1]
@@ -51,7 +51,7 @@ test('return error if input does not meet the criteria', () => {
     const config = noEmpty.process(validatorContext, input, rule)
     expect(config).toBeDefined()
   } catch (e) {
-    expect(e.message).toBe('Input type invalid, expected string as input')
+    expect(e.message).toBe('Input type invalid, expected string or Array as input')
   }
 })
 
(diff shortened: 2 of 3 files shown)


tenshiAMD:
@shine2lay please check if this could be a fix for https://github.com/mergeability/mergeable/issues/654. Thanks in advance.

lib/validators/options_processor/options/no_empty.js
```diff
@@ -26,9 +26,14 @@ class NoEmpty {
     if (typeof input === 'string') {
       isMergeable = input.trim().length !== 0
     } else if (Array.isArray(input)) {
-      isMergeable = input.length !== 0
+      isMergeable = true
```
shine2lay, line 29: I don't understand why this is a `true` case.

lib/validators/options_processor/options/no_empty.js
```diff
@@ -26,9 +26,14 @@ class NoEmpty {
...
     } else {
-      throw new Error(UNKNOWN_INPUT_TYPE_ERROR)
+      if (input == null) {
```
shine2lay, line 31: I think we use lodash `_.isUndefined` would vet the input better to see if its null/undefined

lib/validators/options_processor/options/no_empty.js
```diff
@@ -26,9 +26,14 @@ class NoEmpty {
...
+        // Fallback if `input` is `null` or `undefined`
+        isMergeable = true
```
shine2lay, line 33: if no empty is `enabled`, why should this be true? shouldn't null or undefined be considered `false`?

lib/validators/options_processor/options/no_empty.js
```diff
@@ -25,10 +27,15 @@ class NoEmpty {
 
     if (typeof input === 'string') {
       isMergeable = input.trim().length !== 0
-    } else if (Array.isArray(input)) {
-      isMergeable = input.length !== 0
+    } else if (Array.isArray(input) && input[0] != null) {
```
shine2lay, line 30: I don't understand the changes in these two lines. can you help me understand what you are trying to achieve here?

lib/validators/options_processor/options/no_empty.js
tenshiAMD: @shine2lay thanks for the tip!

lib/validators/options_processor/options/no_empty.js
tenshiAMD: @shine2lay hmm I do not think `_.isUndefined` is enough, I still have to use `== null`

lib/validators/options_processor/options/no_empty.js
tenshiAMD: @shine2lay I'm fixing these lines of tests I've added
```
verify(true, null, [null], 'fail')
verify(true, undefined, [undefined], 'fail')
```

tenshiAMD:
@shine2lay Looks like PR is good to merge now. Please check

lib/validators/options_processor/options/no_empty.js
shine2lay: but why are those tests needed? when it's an array `no_empty` we should just check if the array is not empty i.e `array.length !== 0` . why are we checking for null/undefined value there?

shine2lay: can you describe a use case where that would be useful?

lib/validators/options_processor/options/no_empty.js
```diff
@@ -18,15 +20,17 @@ class NoEmpty {
...
 
     const DEFAULT_SUCCESS_MESSAGE = `The ${validatorContext.name} is not empty`
     if (!description) description = `The ${validatorContext.name} can't be empty`
 
     if (typeof input === 'string') {
       isMergeable = input.trim().length !== 0
     } else if (Array.isArray(input)) {
-      isMergeable = input.length !== 0
+      isMergeable = input[0] != null
+    } else if (input == null || _.isUndefined(input)) {
```
shine2lay, line 32: rather than have it like this, I think it would be cleaner to have 
```
} else if (!(input == null || _.isUndefined(input)) {
throw new Error(UNKNOWN_INPUT_TYPE_ERROR)
} 
```

and remove the last `else`

tenshiAMD: @shine2lay it needs to have the `else if` block so it `null` or `undefined` will be `false` instead of throwing error.

tenshiAMD:
@shine2lay I noticed that there are no testS for input errors. I will add it later when I get online.

tenshiAMD:
@shine2lay All done. Please check. Thanks!

lib/validators/options_processor/options/no_empty.js
shine2lay: you can easily avoid negative conditionals (not (A or B)) if you take the inverse of `or`, instead of `A or B` you can do `not A and not B` that way you can avoid that negation in the front. I personally have no issue either way

lib/validators/options_processor/options/no_empty.js
shine2lay: @tenshiAMD can you please address this comment? I am still not sure why you added these tests where `[null]` is a use case

tenshiAMD:
@shine2lay can you give this a quick review? Thanks a lot.

shine2lay:
@tenshiAMD you haven't address my comment on why you changed the `array.length != 0`

tenshiAMD:
> @tenshiAMD you haven't address my comment on why you changed the `array.length != 0`

@shine2lay ok let me clarify this:

```
[] # this will work
[null] # this will work but it is a false positive
[undefined] # this will work but it is a false positive
```

so I need to change that code

shine2lay:
@shine2lay why is 2nd and 3rd cases false positive? if it's in an array, null and undefined can be used on purpose. Is this something you have already encountered in your use case?

tenshiAMD:
> @shine2lay why is 2nd and 3rd cases false positive? if it's in an array, null and undefined can be used on purpose. Is this something you have already encountered in your use case?

@shine2lay I simply need to pass the tests. You may check the new lines I made in the unit tests.

shine2lay:
@tenshiAMD why are the new lines added in the test? I want to be mindful and not change things that doesn't support a real use case.

tenshiAMD:
> @tenshiAMD why are the new lines added in the test? I want to be mindful and not change things that doesn't support a real use case.

@shine2lay its checking for `null` and `undefined` which are false positive for `no_empty`

shine2lay:
@tenshiAMD I am all for checking null and undefined but the test about checking null inside an array, why inside an array? in what case does this happen?

tenshiAMD:
> I am all for checking null and undefined but the test about checking null inside an array, why inside an array? in what case does this happen?

@shine2lay hmm let's see so if pass `[null]` or `[undefined]` this will return an error not `fail`. Are you good with it?

shine2lay:
@tenshiAMD yes, I am okay with that because there is no obvious use case that adding this is solving.

tenshiAMD:
> @tenshiAMD yes, I am okay with that because there is no obvious use case that adding this feature is solving.

@shine2lay ok then I will remove the new lines in the tests and revert the changes in array check block

shine2lay (approved):
Thanks!

PR #695, opened 2023-02-03
feat: Add Not operator

rm98:
# Goal

Add a `not` operator (for filters and validators) to achieve functional completeness.

This enables each filter or validator to be negated, and is useful for filter/validator options that do not support negation.
For example, one could use `not` before the [`Dependent` validator](https://mergeable.readthedocs.io/en/latest/validators/dependent.html) to validate that certain files were NOT modified in a PR.

# How

I created a new `not` filter and also a `not` validator.
It works exactly like its `and` and `or` counterparts.
Moreover, I renamed `andOrValidatorProcessor` to `logicalConnectiveValidatorProcessor` (for both filter and validator lib) and added the following section to it (shortened):
```js
  if (name === 'Not') {
    if (count !== 1) {
      return consolidateResult([constructError(/* ... */)], context)
    }

    const mapping = {
      error: 'error',
      pass: 'fail',
      fail: 'pass'
    }
    status = mapping[output[0].status]
  }
```

### DIFF
--- a/__tests__/unit/filters/not.test.js
+++ b/__tests__/unit/filters/not.test.js
@@ -1,0 +1,146 @@
+const Not = require('../../../lib/filters/not')
+const Helper = require('../../../__fixtures__/unit/helper')
+
+describe('Not Filter Unit Test', () => {
+  let registry = { filters: new Map(), actions: new Map() }
+
+  beforeEach(() => {
+    registry = { filters: new Map(), actions: new Map() }
+  })
+
+  test('Should pass if subtasks fails', async () => {
+    const not = new Not()
+    const settings = {
+      do: 'not',
+      filter: [
+        {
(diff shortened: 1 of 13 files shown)


hemeroc:
@jusx @shine2lay I would also really appreciate this change could you take a look?

rm98:
Hi @jusx & @shine2lay!

Would appreciate it if you could review this change!

shine2lay (approved):
LGTM

shine2lay:
will merge it once CI passes

shine2lay:
It seems lint is failing, please fix

rm98:
Hi @shine2lay - thanks for checking!
I fixed the linting error.

shine2lay:
@rm98 thanks for the fix

PR #696, opened 2023-02-03
feat: Add team option to author filter

rm98:
# Goal

Add a way to change behavior based on team membership.

This PR adds the `team` option to the `author` filter.
If the `team` is set, then it is verified that the `author` is part of the team, otherwise, the filter fails.

Together with #695, this enables checking whether an author is _not_ part of a team.

# How

I added `team: 'string'` to the supported settings of the author filter.
Then, I added a `TeamProcessor` that uses the existing functionality for `Teams` from the `validators` to check whether the user is part of the team.


rm98:
Hi @shine2lay - this is another small feature, I rebased it on top of master after the previous PR was merged 🙂

shine2lay (approved):

PR #705, opened 2023-03-13
fix: Replace delete with remove in changeset validator

dylanslavin:
This validator doesn't work as expected, and according to [Github's API docs](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files), the `status` field of a file can be `removed`, but not `deleted`.

Edit: link


jusx (approved):

dylanslavin:
It looks like the CI build failed when this PR was merged, and the new behavior isn't reflected in production. Would you guys mind re-running the failed build?

shine2lay:
@dylanslavin I re-triggered the CI, it's passing now. The changes should be reflected within a few mins

PR #710, opened 2023-04-25
feat: Add author validator

rm98:
# Goal

Add an `author` _validator_ that behaves exactly like the `author` _filter_.

# How

I created `lib/validators/author.js`, which is basically the same as `lib/filters/author.js`. I saw no convention on how to do an object that can be a filter and validator at the same time.

I also created `lib/validators/options_processor/options/team.js`, which is pretty much what `lib/filters/options_processor/options/team.js` is. However, there is a subtle difference when implementing it: the method signature for `process` contains an extra argument for `filters`, thus, we cannot reuse the `TeamProcessor`.

PR #713, opened 2023-05-10
fix: Loading teams for team option of author filter/validator

rm98:
Fix loading teams for team option of author filter/validator.

## Context

My previous PR #710 added an author validator. However, the API classes (mocked in the tests) access the context, and the context in `lib/validators/options_processor/options/team.js` is not the one I expected (it's a special validator context, different from the context in `lib/filters/options_processor/options/team.js` (correct one).

API calls are mocked for unit tests, but they log information taken from the context, and this doesn't work with the validator context.

## Changes

I introduced a new function in `lib/validators/options_processor/teams.js` and moved most of the logic there (shared for filter/validator), and removed the `options_processor/options/team.js` classes because we don't always have the _correct_ context there.
This method is called prior to processing the other options (and potentially results in an early return if unsuccessful), and then the `team` key is simply removed from the options. 

This follows the pattern for similar cases, for example, see [lib/validators/changeset.js#L58](https://github.com/mergeability/mergeable/blob/27dd124799c4fa2fb7436b0484ece860132d4e49/lib/validators/changeset.js#L58)


rm98:
@shine2lay could you also check out this PR? Thx!

shine2lay (approved):

Issue #711, opened 2023-05-02
[BUG] - Getting an error for a specific PR


oferprtz-atbay:
Hey,

Using mergeable for specific repo and all PR's are working correctly.
today ive created new PR with just helm files changes and got this error from mergeable:
```
Error Details
HttpError: Invalid request.

No subschema in "anyOf" matched.
For 'anyOf/0', {"labels"=>{"labels"=>["Ready for Review"]}} is not an array.
For 'anyOf/1', {"labels"=>{"labels"=>["Ready for Review"]}} is not an array.
For 'properties/labels', {"labels"=>["Ready for Review"]} is not an array.
```

on the checks tab im getting this:
![image](https://user-images.githubusercontent.com/125448935/235739388-50a2fe8e-49ac-4a5e-8a67-ac6df23f5124.png)

Didnt change mergeable yaml for a while now.

thanks,
Ofer

oferprtz-atbay:
Hey @shine2lay thanks for the quick reply.
when creating the PR the mergeable start to run and post this:
![image](https://user-images.githubusercontent.com/125448935/235858051-07cb44ec-132a-449e-81d7-d02f3d6bf9e2.png)

expected result:
![image](https://user-images.githubusercontent.com/125448935/235858177-087f67e9-c74d-473f-b6e2-fc7e2816c349.png)

not sure why its happening, the mergeable config runs on all other repos smoothly.
here is the config:
`version: 2
mergeable:
  - when: schedule.repository
    validate:
      - do: stale
        days: 30
        type: pull_request
    pass:
      - do: comment
        payload:
          body: This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 10 days.
      - do: labels
        add: 'Stale'
          
  - when: pull_request.*
    validate:          
      - do: description
        no_empty:
          enabled: true
          message: Description matter and should not be empty. Provide detail with **what** was changed, **why** it was changed, and **how** it was changed.
      - do: or
        validate:
          - do: title
            must_include:
              regex: '\[(DEVO|RND|CP|DP|DS|CR)-\d+\]|\[SRE\]|\[DEVOPS\]'          
    pass:
      - do: labels
        add: 'Ready for Review'
      - do: checks
        status: 'success'`

shine2lay:
I have never seen this before, any chance the this could be complaining about the label not existing?

oferprtz-atbay:
[...]

PR #715, opened 2023-05-11
fix: Send correct payload for changing labels

rm98:
See discussion in #711


shine2lay (approved):
Thanks for the fix

shine2lay:
@rm98 i'll go ahead and merge it, it should take affect within minutes, please ping me if the error still persist.

Issue #597, opened 2021-11-13
Release & publish docker images with versioning after release is completed


quike:
We are coming from https://github.com/mergeability/mergeable/issues/554

The purpose of the issue would be to build and publish a new docker  mergeable image with the new version after the release is completed.
Currently the docker job is ignored and no new builds are published. The idea is to connect both workflows so one follows another one and new artifacts can be deployed to be consumed with the right version.

Besides current Dockerfile is off in comparison of nodejs version used by the app 

Dockerfile: `FROM node:12.16.0`
package.json: `14.15.4`

@shine2lay any reason why docker images are not built? github rate limitations?

shine2lay:
@quike i am not aware of a blocker to build docker images, we just haven't gotten around to it.

boekkooi-lengoo:
@shine2lay are you planning to enable pushing to https://hub.docker.com/r/mergeability/mergeable?
It would be great to have an official build. 
Based on [.github/workflows/docker.yml](https://github.com/mergeability/mergeable/blob/master/.github/workflows/docker.yml) it looks like [adding the secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository) `DOCKERHUB_PUSH`, `DOCKERHUB_USERNAME`, `DOCKERHUB_PASSWORD` and `DOCKERHUB_REPO` would enable this.

tenshiAMD:
@boekkooi-lengoo looks like it is purposely disabled by them

Ch4s3r:
The issue linked in the `docker.yml` regarding the [workaround](https://github.community/t5/GitHub-Actions/jobs-lt-job-id-gt-if-does-not-work-with-env-secrets/m-p/38549) does not exist anymore.
Is there a possibility to enable the publishing of docker images again?

PR #751, opened 2024-06-07
feat: publish multi arch docker image to dockerhub

Ch4s3r:
When a new release tag is created, trigger a new docker image build, using the provided `Dockerfile`.
Builds an image for `arm64`, `amd64` and `arm/v7` on parallel running runners with different architecture, for native compilation speed.
Combines all images into one multi arch build and uploads it to dockerhub as `mergeability/mergeable`.
This image is then tagged with the version from the github tag and the `latest` tag.
Can also be triggered manually which then only creates the `latest` tag.

Most of the code comes from the official [docker docs](https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners) on how to build a multi arch image.

Fixes https://github.com/mergeability/mergeable/issues/597

Tested and verified it works when pushing a release tag.

image

Todos:
- [x] Change`secrets.DOCKERHUB_PASSWORD` to `secrets.DOCKERHUB_TOKEN` (not sure if password is still accepted by dockerhub)


Ch4s3r:
@shine2lay could you please have a look when you have time ? :)

shine2lay (approved):

shine2lay:
@Ch4s3r I checked that we have DOCKERHUB_TOKEN in repository secrets

PR #752, opened 2024-06-09
docs: corrections and completions and alignments

andekande:
improving docs having stumpled upon some problems during self-hosting and extensive usage


shine2lay (approved):

shine2lay:
@andekande thanks

PR #753, opened 2024-06-10
fix: Docker image not working

Ch4s3r:
The old `Dockerfile` was using nodejs version 16 and not building at all.
This PR adds a new setup created with `docker init`.
It uses node 18 now as stated in the `package.json`.

These versions need to be kept in sync, so updating the nodejs version must also happen in the `Dockerfile`.

Executing the new docker image works:

image


Ch4s3r:
@shine2lay @jusx please have a look :)

shine2lay (approved):

PR #764, opened 2024-06-25
feat: Add buildpacks for building docker image

Ch4s3r:
- Use semantic-release github action
- Use buildpack for optimized docker images
	- Only build amd64 image as arm is not supported by now from paketo buildpacks
	- smaller image size 350 vs 100mb now
	- publishes `latest` and versioned tag
- Remove unnecessary matrix build from test
- Add config to `package.json` that the package is public (as the github action fails to publish to npmjs otherwise with 402)

npmjs:
image
docker:
image

- [x] Test that merging PR creates new version
- [x] Add changelog entry


Ch4s3r:
Always get this error:

```
An error occurred while running semantic-release: Error: Command failed with exit code 128: git tag v2.22.0 fcbbbb3152e4be8453cb4d03d8f8ad50f94cab4c
fatal: tag 'v2.22.0' already exists
```

Need to check how to skip it or why the tag is there.

.github/workflows/testing.yml
```diff
@@ -1,31 +1,16 @@
...
 
-on:
-  push:
-    branches: ['**']
-  pull_request:
-    branches: ['**']
-
-permissions:
-  contents: read
+on: [ push ]
```
Ch4s3r, line 3: Was there any reason why two trigger have been used?
As in the last example [here](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs#specifying-the-nodejs-version) they do it via `push`.

Ch4s3r:
@shine2lay @jusx please have a look :)

hemeroc:
Hi, that would be real cool to have.
@shine2lay @jusx is there any blocker for this?

shine2lay (approved):

thewitcher-777-cloud (approved):

